PR #65, opened 2015-01-02
Use new array syntax

crumblingstatue:


jeremyletang:
Hey thanks for the fix !

PR #66, opened 2015-01-04
Build & Repo Enhancements

eriknelson:
- Add optional LDFLAGS to build to support custom installation location
  of SFML/CSFML dependent libraries.

- rm build products on clean

- Ignore build products in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 doc
 target
+bin/
+lib/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ doc
 target
 bin/
 lib/
+Cargo.lock
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ name = "rust-sfml"
 description = "Rust binding for sfml"
 version = "0.9.0"
 authors = ["Jeremy Letang <letang.jeremy@gmail.com>"]
+build = "build.rs"
 
 [lib]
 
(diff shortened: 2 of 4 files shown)


eriknelson:
Looks like the build failure is due to the recent nightly, I'm on the following nightly with no issue:
rustc: 39d740266 2015-01-01
cargo: 1a1868b 2014-12-31

eriknelson:
Addresses #67

jeremyletang:
Hey, thanks for the changes !

Do you use the makefile to build rsfml ? I've let it in the repo when I switched to cargo, but I planned to remove it soon.

erik-sb:
Hey Jeremy,

I think this should cover both the cargo build and the make build. Personally, I prefer to use cargo if possible since it's becoming more idiomatic in the rust ecosystem. The motivation behind this was because the distro I'm using did not package the required C/SFML libs. I had to build them from source, which I prefer to make install to a non-sudo prefix location to isolate them from the package manager installed libs.

eriknelson:
Hey Jeremy,

I think this should cover both the cargo build and the make build. Personally, I prefer to use cargo if possible since it's becoming more idiomatic in the rust ecosystem. The motivation behind this was because the distro I'm using did not package the required C/SFML libs. I had to build them from source, which I prefer to make install to a non-sudo prefix location to isolate them from the package manager installed libs.

Syntaf:
Is there any movement towards getting this merged? [ref](http://stackoverflow.com/questions/31630064/rust-sfml-linking-with-cc-failed-how-to-specify-lib-include-share-path/31630169#31630169)

PR #70, opened 2015-01-08
Update to run with latest Rust Nightliy. Mostly new c_string design

sindreij:
Hello

I have tried to update rust-sfml to be able to build with the newest Rust Nightly. This include the pull-request from crumblingstatue.

The c_string redesign meant I had to change lots of methods. I am not sure that what I have done is the best, but it is to my best understanding of how to do this. I also have not been able to test all the methods I have changed. There are a lot of methods, and I do not even know how every one should behave. However, all the examples run.

I hope this helps :)

PR #71, opened 2015-01-11
Update to the last rust-nightly.

sebcrozet:
This does not fix the warnings.


sebcrozet:
Btw, this includes #70.

jeremyletang:
thanks for the fixes !

PR #72, opened 2015-01-13
update to last Rust

jeremyletang:


x4nt:
You should drop the 'u' suffix altogether. Instead of '6us', apparently '6' is preferred.

PR #75, opened 2015-01-14
Updating a texture from a window now takes the window by reference

Bastacyclop:


jeremyletang:
Thanks ! it was fast !

PR #76, opened 2015-02-07
Update to latest Rust by switching to 'old_io' in sound capture example.

pshendry:
Figured this would be more stable than trying to use the new std::io (AFAIK it's still in flux).


pshendry:
Closing and re-opening in order to re-run that build, which failed while trying to get the Rust nightly.

PR #79, opened 2015-03-25
Update to latest Rust

crumblingstatue:


crumblingstatue:
The build fails because the travis build is invoking `rustdoc` instead of `cargo doc`, making it think we are using built-in libc, when we are actually using libc from crates.io.

crumblingstatue:
With this, rust-sfml should build on rust 1.0.0-beta.

jeremyletang:
@crumblingstatue thanks for the updates !

Issue #80, opened 2015-04-13
RFC: Rename crate from `rsfml` to `sfml`


crumblingstatue:
The crate is already used in the context of Rust, so the `r` is redundant.
`sfml` is one letter shorter, and more consistent with other bindings like rust-sdl and rust-sdl2,  which use `sdl` and `sdl2` for their crate names respectively.

The only drawback is that existing code would need to be changed to use `sfml` instead of `rsfml`, but that sounds like just a simple search&replace operation.

Alternatively, we could use `sf`, which is even shorter, and in line with the original SFML library, which is contained in the `sf` namespace.

jeremyletang:
Okay. let's do this change.

crumblingstatue:
Implemented in 61685f2c0094def9620f392cf5220b02e23d8f31.

PR #81, opened 2015-04-14
Rename package from rust-sfml to sfml

crumblingstatue:
I left this out of #80 as an oversight.

Basically the same arguments apply. The package is used in the context
of Rust, so the `rust-` prefix is redundant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 
-name = "rust-sfml"
+name = "sfml"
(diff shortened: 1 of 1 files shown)


crumblingstatue:
@jeremyletang 
What do you think?

jeremyletang:
@crumblingstatue I'm okay with this.

( Have not see before as I'm not notified for each your push on the repository but you make a lot of improvement, great job !).

PR #85, opened 2015-04-18
Rename non_snake_case public methods to snake_case

crumblingstatue:
Changes public methods of Color and RenderTarget, which used UPPERCASE for acronyms.

I think this is appropriate if we want to be compliant with Rust naming conventions, but I want to know what @jeremyletang thinks before renaming them.


jeremyletang:
@crumblingstatue Ok for me. Let's merge this.

PR #88, opened 2015-04-19
graphics::vertex_array: Bind Vertices' lifetime to VertexArray

### LINKED ISSUE #87: Lifetime in `Iterator` impl for `Vertices` is not constrained
https://github.com/jeremyletang/rust-sfml/blob/1c91c6371f674ec044c94fb84a4db24ee9dd148f/src/graphics/vertex_array.rs#L256 does not compile on latest nightly:

```
src/graphics/vertex_array.rs:256:6: 256:8 error: the lifetime parameter `'s` is not constrained by the impl trait, self type, or predicates [E0207]
src/graphics/vertex_array.rs:256 impl<'s> Iterator for Vertices {
                                      ^~
```

I talked to people on #rust, and they are telling me that it was a bug that this compiled before, and this code is memory-unsafe.

We must either constrain the lifetime in a way to be memory-safe, or return raw pointers instead of references.

/cc @jeremyletang

crumblingstatue:
Closes #87.

/cc @jeremyletang

### DIFF
--- a/src/graphics/vertex_array.rs
+++ b/src/graphics/vertex_array.rs
@@ -39,8 +39,8 @@ pub struct VertexArray {
 }
 
 /// An iterator over the vertice of a VertexArray
-pub struct Vertices {
-    vertex_array: *mut ffi::sfVertexArray,
+pub struct Vertices<'a> {
+    vertex_array: &'a VertexArray,
     pos: u32
 }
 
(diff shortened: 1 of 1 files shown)


crumblingstatue:
I'm going to merge this because rust-sfml can't build on nightly without it.
If it turns out to be incorrect, we can always fix it later.

Issue #94, opened 2015-05-19
2.3 update tracking


jeremyletang:
Here we will list the different item who have change since `SFML 2.1` and that we need to update for the 2.3 release:
* Graphics
 - [ ] BlendMode
 - [ ] CircleShape
 - [ ] Color
 - [ ] ConvexShape
 - [ ] Font
 - [ ] FontInfo
 - [ ] Glyph
 - [ ] RectangleShape
 - [ ] RenderTexture
 - [ ] RenderWindow
 - [ ] Shape
 - [ ] Text/TextStyle
 - [ ] VertexArray
* Window

brandonson:
This could really use a look from someone - you get segfaults with even simple shape drawing if you try to use this with SFML 2.3.  I may be able to look at making things work in 2 or 3 weeks, but if someone gets to it first that would be great.

Ketsuban:
It's been six months since the last comment on this issue. The version available on crates.io (i.e. what people are going to use) hasn't been updated since *April*. The [advertised version was bumped to 2.3](https://github.com/jeremyletang/rust-sfml/commit/226ccdf57edf7685b5bbd701645d0338ed51668d) in October, but none of the items in the checklist are checked. If the version mismatch has been resolved, could you close this issue and push a version bump to crates.io?

crumblingstatue:
Only *some* work has been done towards updating to 2.3.

The version bumped was the version that the master branch works on, which is indeed 2.3. Note that the majority of the API wasn't updated to 2.3, but some of the internals have been reworked not to crash with SFML 2.3, which in turn made it incompatible with versions < 2.3.

All this being said, I could still release a new version on crates.io, just note that it doesn't fully support 2.3.

crumblingstatue:
Actually, the package won't publish as-is.

The project now consists of multiple cargo packages, and some more work will have to be done before a new version can be published.

crumblingstatue:
Alright, I published [0.10.1](https://github.com/jeremyletang/rust-sfml/releases/tag/v0.10.1).

Ketsuban:
Thanks. :)

Scellow:
Wouldn't it be better to fork CSFML, and then start to implement their commits one by one by pulling the commits, so you'll get a track of each change, in the long term it'll be easier to manage, and it'll be easier for other people to contribute since they'll know exactly what's missing

crumblingstatue:
[...]

Issue #93, opened 2015-05-19
maybe replace IntRect / FloatRect to to Rect<T>


jeremyletang:
This will reflect the recent change on `Vector2f/Vector2i/Vector2u/Vector2<T>`.
The difference is that IntRect / FloatRect have two methods: `contains and intersect`. We may need to reeimplement them.

Issue #92, opened 2015-05-19
maybe use bit flags for enums


jeremyletang:
We may need to use bitflags for our enums, or at least implement `BitOr, BitAnd, BitXor` if it can be safe enough.

Issue #31, opened 2014-03-02
Reason for refcounted views?


aepsil0n:
In `rsfml::graphics::RenderWindow` all the methods dealing with views (`set_view`, `get_view` …) deal with Rc<RefCell<View>> objects in their instances. I wonder why this is. It feels a bit bloated to deal with these objects instead of simple borrowed pointers. So in essence my question is the following: why should the RenderWindow care at all about how the view's memory is managed?

Simply want to understand your rationale behind this.

crumblingstatue:
I don't think this applies anymore. `Rc` is nowhere to be found in the `render_window` module anymore.

PR #97, opened 2015-05-20
Massive Overhaul and Improvements

SpaceManiac:
I have been working for some time working on a massive set of improvements and updates to the library. I wanted to hold off on submitting a pull request until I stabilized things a little more and re-added some of the functionality I destroyed to make things easier, but it looks like you're back to working on the library in earnest, so I figured I'd open this to try to prevent wasted effort.

This pull request isn't really mergeable as-is: it conflicts with many of your most recent changes, some functionality like the network module and refcounted resources are unavailable, and there's still improvements to be made. It does partially or completely fix many open issues, however, so I would like to bring the changes up for discussion before too much effort is duplicated.

I can understand if some of these changes conflict with your vision for the binding, but I wanted to open things up to discussion. If you want to more easily see an overview of the structure as I have built, I recommend using `cargo doc`. If needed, I can host a copy of these docs.

Breakages:

* The `network` module is missing.
* Reference-counted versions of `Sprite`, `Sound`, etc. are missing.
* Backwards compatibility is completely shot.

Changes and fixes (see the commit list for specifics):

* #31 - view management returns properly borrow-checked references.
* #92 - bit flags are used for TextStyle and WindowStyle.
* #93 - IntRect/FloatRect are `Rect<T>`.
* #94 (partial) - entirely updated to CSFML 2.2, pending release of CSFML 2.3.
* Replaced `system` module with a native Rust implementation, to remove the
  `libcsfml-system` dependency.
* Replaced VertexArray with a native Rust implementation wrapping a `Vec<Vertex>`.
* Removed re-exports of enum variants - they are instead scoped to their enum.
* Changed mouse and keyboard functions to be methods on `MouseButton` and `Key`.
* Refactored a lot of incredibly repetitious or otherwise ugly code (such as matches on `SfBool`).
* Reduced `Drawawble` to one method (matching C++) and made it object-safe.
* Added `Foreign` wrapper type to simplify ownership and mutability correctness.
* Heavily tidied `ffi::` modules.
* Fixed broken Unicode handling on Window titles and Text strings.
* Added `Transformable`, `Shape`, `SoundSource`, other traits representing common functionality.
* Changed `rand` crate to only a dev-dependency, since it's only used by examples.
* Implemented `Default`, `Debug`, etc. on more applicable types.
* Added construction from memory and from input streams to all resources.
* Added `SoundRecorder` and `SoundStream` for custom audio input and output.
* Removed unsound `Events` iterator types.
* Other miscellaneous missing functionality.
* Updated all the examples to reflect the changes.
* Reviewed the entire documentation and fixed all problems found.

To do:

* Re-add the `network` module.
* Re-add ref-counted versions of things (possibly with generics).
* Improve ergonomics of Vector2 and Vector3 conversion and use.
  * Potentially replace `*2f` methods with either the use of `(x, y).into()` or
    generic methods taking `<T: Into<Vector2>>`.
* Add more checks to prevent undefined calls to SFML functions.
* Improve documentation regarding error conditions.
* Possible other internal improvements.
* Help investigate Cargo and Travis configuration.
* Implementation of `Clock` for OSX.
* Fix the incredibly broken indentation (sorry).
* Problems pending CSFML changes:
  * Non-window-specific mouse and touch positioning?
  * `keyboard::set_virtual_keyboard_visible`
  * Missing methods provided only for `SoundRecorder` but not `SoundBufferRecorder`
* Missing license headers on some files.
* Update README.

Thanks for your time.


jeremyletang:
Hey, Thanks for taking time to improve rust-sfml.

I will be happy to merge your improvement in the repository. But It will really improve le process if you can create different pull request for each different feature. For now It's really difficult to review the code.
Moreover open different pull request  issue will better to talk about the change as there is some I'm not really okay to merge.

SpaceManiac:
Sounds fine. I can work on getting the smaller features or cleanup into separate pull requests, but are there any specifically you'd prefer to see first, or ones you're less interested in?

jeremyletang:
Okay, I think we can begin with the code cleanup you've made, then the code refactor (like Foreign, etc), then continue with missing feature from the binding (SoundRecorder, etc), and finally with the port 2.3.

There is some thing really less interested in but we can discuss about:
* removing of the libcsfml-system dependency, I think it's not a problem to be dependent of it even fo som feature, and I'm not really okay to rewrite functionnality in full Rust, if we use the original lib we are sure users will have the expected behaviour, and we don't have to check at each update if a change is made inside the original implementation.
* Same for VertexArray, same reason.
* Why would you remove Events iterator type, it seems pretty useful I think ?
* I'm not sur we need methods for Key and Mouse, standalone function seems sufficient.

SpaceManiac:
`system`: my main justifications are that aside from the one function acting as the `Clock` implementation, everything is trivial math and could benefit from optimization in the absence of an FFI call, and that on Windows it means one less .dll to deal with (admittedly it's a pretty small one).

`VertexArray`: the main argument is ergonomic. In C++ SFML, a `VertexArray` is a trivial wrapper around a `std::vector<Vertex>` and a `PrimitiveType` that exposes an annoyingly small interface instead of just allowing access to the `vector`. IMO, it's easier to work with a `VertexArray` when it's more directly usable like any old `Vec<Vertex>`, and a user of the library won't have to learn even more new API to make use of it. The same argument applies to `ConvexShape`.

The problem with the Events iterator type is more complex. As it stands (containing a `*mut sfWindow`), it's memory unsafe - the compiler will allow the Events to outlive the Window from which it was created, which means bad things if it's used after the Window is destroyed. If Events is instead changed to have a lifetime and refer to the Window from which it was created (making it memory-safe), it becomes less useful: it mutably borrows the Window, so code like this will fail to compile:
```rust
for event in window.events() {
    match event {
        Event::Closed => window.close(),
        _ => {}
    }
}
```
Without some other solution, the Events iterator is either unsafe or useless. The `poll_event` solution is less pretty, of course, but at least it works:
```rust
while let Some(event) = window.poll_event() {
    match event {
        Event::Closed => window.close(),
        _ => {}
    }
}
```

The Key/Mouse methods are mainly to prevent the need to separately import the free-standing functions, and for general ergonomics (`Key::X.is_pressed()` vs `keyboard::is_key_pressed(Key::X)`).

jeremyletang:
Okay. I stay not convince for sfml-system. For sure we already have rewrite the vector part, for the others, mainly the clock, I cannot test it on windows when I know that the sfml version work, moreover, depends on this library don't seems as a problem as by default this is installed when you install sfml/csfml.

For the VertexArray/ConvexShape I will see your implementation before to think about.

For the Events iterator, I really like this , event if it's a bit unsafe, there is no real reason to destroy the window when you iterate the events, and we can check the window is not null inside the event.

About the Key/Mouse, this use seems really more ergonomic for sure.

SpaceManiac:
I am willing to cede on sfml-system. As you say, it's a small and obtained-with-the-rest dependency, and it is a burden to make sure a reimplementation is correct.

As much as the Events iterator is nice, if it's unsafe it's fundamentally pointless. Rust as a language is all about maintaining memory safety - if the Events iterator leaks memory safety without being marked `unsafe` (and thus being quite unergonomic to use), it violates the expectations that the compiler has that `unsafe` blocks maintain Rust's invariants and that the user has that the compiler will be able to catch all memory safety problems - even dumb ones, like using Events after the window is freed:
```rust
let events = Window::new(...).events();
let event = events.next();
```
Of course this is terrible code, but it will segfault if we let it, and the whole point of Rust is to catch this kind of thing at compile time.

crumblingstatue:
This pull request is too massive, and accumulated too much merge conflicts to be considered for merging.

However, there are a lot of great ideas here that we should consider individually.

PR #98, opened 2015-05-22
Cleaned up manual matching on SfBool.

SpaceManiac:
Instead, methods `SfBool::from_bool` and `to_bool` are used to remove the heavy duplication that was present beforehand.

One possible improvement that could be made is use of the `From` and `Into` traits to avoid the need to import `SfBool` anywhere at all.

From #97.


jeremyletang:
hey thanks, this clean the code a lot already !

PR #99, opened 2015-05-22
Turned IntRect and FloatRect into Rect<T>.

### LINKED ISSUE #93: maybe replace IntRect / FloatRect to to Rect<T>

SpaceManiac:
Native Rust implementations of the methods on `Rect<T>` are used to allow for types other than `i32` and `f32` to be used by client code.

From #97. Fixes #93.

### DIFF
--- a/src/graphics/rect.rs
+++ b/src/graphics/rect.rs
@@ -22,49 +22,32 @@
 * 3. This notice may not be removed or altered from any source distribution.
 */
 
-//! Manipulating 2D rectangles
-//!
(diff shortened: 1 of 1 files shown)


jeremyletang:
This cannot be merge, there is conflict with #98

SpaceManiac:
Whoops. Fixed.

PR #103, opened 2015-05-23
Replaced Color::add and modulate with Add and Mul.

SpaceManiac:
The previous Mul implementation was broken, and both add and modulate can be
better expressed as operators.

Also includes documentation cleanup on `Color` (open to discussion).

From #97.

### DIFF
--- a/src/graphics/color.rs
+++ b/src/graphics/color.rs
@@ -74,28 +74,6 @@ impl Color {
         })
     }
 
-    /// Add two colors
-    ///
-    /// # Arguments
-    /// * color1 - The first color
-    /// * color2 - The second color
-    ///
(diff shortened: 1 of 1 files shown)


SpaceManiac:
I fixed the merge conflicts and simplified the diff to exclude the extraneous docs changes. I argue that it doesn't make sense to supply the following set of operations:

* `add` method that calls to FFI
* `modulate` method that calls to FFI
* `Add` impl that re-implements `add` in Rust
* `Mul` impl that re-implements `modulate` in Rust **incorrectly**

when the following set of operations is sufficient:

* `Add` impl that calls to FFI
* `Mul` impl that calls to FFI

PR #104, opened 2015-05-23
Changed is_key_pressed and is_button_pressed to be methods.

SpaceManiac:
Also tidied documentation on `Key` and `MouseButton`.

This improves ergonomics, changing:
```rust
keyboard::is_key_pressed(Key::X)
mouse::is_button_pressed(MouseButton::Left)
```
to
```rust
Key::X.is_pressed()
MouseButton::Left.is_pressed()
```

From #97.

### DIFF
--- a/src/window/keyboard.rs
+++ b/src/window/keyboard.rs
@@ -22,18 +22,15 @@
 * 3. This notice may not be removed or altered from any source distribution.
 */
 
-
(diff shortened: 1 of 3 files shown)


jeremyletang:
hum, I've just merged but i think that some examples should probably not build anymore, we need to fix this.

PR #107, opened 2015-08-16
Make system::Time copyable.

brandonson:
Not having this is annoying when trying to do something like
`current_elapsed - last_update > update_interval` to check
if things need an update.


crumblingstatue:
Make sense to me, since the underlying `ffi::sfTime` is also `Copy`.

PR #108, opened 2015-08-19
Make Drawable::draw callable on trait objects

emlai:
Now you can f.ex. store a `Vec` of `Drawable`s (trait objects), loop
through them and call `draw` on each of them.

More information about this issue:
http://stackoverflow.com/q/32081065/3425536

### DIFF
--- a/examples/custom_drawable.rs
+++ b/examples/custom_drawable.rs
@@ -31,7 +31,7 @@ impl<'s> CustomDrawable<'s> {
 
 // Implements the drawable trait, only this function is mendatory.
 impl<'s> Drawable for CustomDrawable<'s> {
-    fn draw<RT: RenderTarget>(&self, render_target: &mut RT, _: &mut RenderStates) {
+    fn draw(&self, render_target: &mut RenderTarget, _: &mut RenderStates) {
         render_target.draw(&self.circle);
         render_target.draw(&self.rect)
     }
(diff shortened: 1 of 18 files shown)


emlai:
Fixing build…

crumblingstatue:
This replaces static dispatch by dynamic dispatch, I'm not sure what @jeremyletang would think about this, so I'll wait for him to review it.

jeremyletang:
Hey,
I'm a bit annoyed with replacing static dispatch with dynamic dispatch, I don't know the runtime cost on a collection of a lot of objects.
Is it possible to add the method to the trait and keep the existing one and keep the object safe using this method ? Or all the methods should respect objects safety rules ?

emlai:
> Is it possible to add the method to the trait and keep the existing one and keep the object safe using this method ? Or all the methods should respect objects safety rules ?

I have no idea, unfortunately.

If you think static dispatch is better in this case then you can close this PR, as far as I'm concerned.

crumblingstatue:
We will probably want to decide whether to support dynamic dispatch or not before 1.0.

crumblingstatue:
We might want to consider supporting this, because the C++ SFML API also supports dynamic dispatch of `RenderTarget`.

crumblingstatue:
Closing in favor of #149.

Issue #73, opened 2015-01-13
About an update to handle cargo build script


Noxivs:
Hi,
following [this](http://www.reddit.com/r/rust/comments/2s9261/how_do_i_build_a_sdl2_project_on_windows_where_do/cnnftga) reddit subject. 

The ffi mod should be outside the rust sfml "wrapper" as an internal project to allow an overriding of the build script.

More conventional for cargo and easiest way to link for windows users.

There are a lot of things to move out and I am not sure of what to keep. But it could look like that

Cargo.toml
```
[package]

name = "sfml-sys"
version = "0.0.1"
authors = ["Jeremy Letang <letang.jeremy@gmail.com>"]
links = "sfml-sys"
build = "build.rs"

[lib]
name = "sfml-sys"
path = "src/lib.rs"

[build-dependencies]
pkg-config = "*"

[features]

default = []
use-pkgconfig = []
```

build.rs
```
extern crate "pkg-config" as pkg_config;

fn main() {
    let libs = [
        "csfml-system", 
        "csfml-window", 
        "csfml-graphics",
        "csfml-audio",
        "csfml-network"
    ];

    if std::os::getenv("CARGO_FEATURE_USE_PKGCONFIG").is_some() {
        for &l in libs.iter() {
            if pkg_config::find_library(l).is_err() {
                panic!("Could not find {:} via pkgconfig")
            }
        };
    } else {
        for &l in libs.iter() {
            println!("cargo:rustc-flags=-l {:}", l);
        };
    }
}
```
Regards.

PR #109, opened 2015-09-24
Move ffi bindings to separate -sys packages

crumblingstatue:
Following [Cargo conventions](http://doc.crates.io/build-script.html#-sys-packages), move bindings to each native csfml library into their own -sys package.

This has the following advantages, among possibly others:
- Clearly separated, reusable low level bindings to the various CSFML bindings, which could be utilized by projects other than rust-sfml.
- Easily configurable linking to CSFML, allowing to link to CSFML installed in non-standard locations, etc (Close #66 and #73).

This allows for, possibly in the future, splitting rust-sfml into packages for each sfml module, allowing projects to only use the modules they need (e.g. a project that only wants to play audio would only need to depend on the `sfml-audio` package).

The work so far has been focusing on moving the declarations out to their own packages and making  rust-sfml compile with the changes. There are some problems that I don't know how to best solve, for example guidelines tell that -sys crates shouldn't define functionality (e.g. `impl` for primitive types), so I have made newtype wrappers for some of these types. There is also the problem with csfml using concrete types, e.g. `sfVector2i`, while rust-sfml has templated types (`Vector2<T>`). I'm not sure if I should put the templated version inside the -sys bindings. Right now, both concrete `sfVector2` variants defined in the -sys bindings and templated `Vector2<T>` defined in the high level bindings exist, and I transmute between them. This is not ideal.

cc @jeremyletang


crumblingstatue:
I'm going to wait a few days and keep working on it. If by then there are no objections, I will merge it into master.

jeremyletang:
hi @crumblingstatue,

This seems to be really great stuff. About the declaration of the parametrized `Vector2<T>`, I know the guidelines don't tell to move them inside the `-sys`. But in this case it's almost a base type from the binding, which is not bind, so i think we can make the declaration inside the `-sys` crate then re-export them inside the library, then we don't have to transmute inside the library.

About the fact of splitting the binding in different sub library, I'm bot sure it's the best way to use it. For sure this will match the c++ way the library is provided, but each library is at least dependent of `csfml-system`, and for the case of `csfml_graphics`, it depends of two library `csfml-system` and `csfml_windows`. I think it's easier to link with only one dependency than specify each of them.

crumblingstatue:
> About the declaration of the parametrized Vector2<T>, I know the guidelines don't tell to move them inside the -sys. But in this case it's almost a base type from the binding, which is not bind, so i think we can make the declaration inside the -sys crate then re-export them inside the library, then we don't have to transmute inside the library.

I suppose you're right, although then we also have to provide the trait implementations inside the -sys crate as well, due to coherence rules.

> About the fact of splitting the binding in different sub library, I'm bot sure it's the best way to use it. For sure this will match the c++ way the library is provided, but each library is at least dependent of csfml-system, and for the case of csfml_graphics, it depends of two library csfml-system and csfml_windows. I think it's easier to link with only one dependency than specify each of them.

There could be individual crates for those who want to control what they link to, and an `sfml` crate that contains all of the modules.

crumblingstatue:
Rebased.

crumblingstatue:
@jeremyletang I have moved the generic type definitions into a crate called `sfml-types`, which the rest of the -sys crates depend on. Since these types are repr(C), I assume they should interface with C without problems, even if they are generic.

Anyway, I am going to merge this now, as there doesn't seem to be any resistance against this change, and I want to start working on updating the bindings to SFML 2.3.

If any problems surface, we can fix them later.

PR #113, opened 2016-03-03
Making bitflag-type enums public for reexporting.

Rafagd:
Stable Rust couldn't compile the enums bellow because they were private.

Enums changed are:
- graphics::text_style::TextStyle
- window::window_style::WindowStyle

### DIFF
--- a/src/graphics/text_style.rs
+++ b/src/graphics/text_style.rs
@@ -29,7 +29,7 @@
 bitflags! {
     #[doc="Available text styles."]
     #[repr(C)]
-    flags TextStyle: u32 {
+    pub flags TextStyle: u32 {
         #[doc="Regular characters, no style."]
         const REGULAR = 0,
         #[doc="Bold characters."]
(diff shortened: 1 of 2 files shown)


crumblingstatue:
Out of curiosity, what rustc version are you using?

rust-sfml compiles on 1.6.0 stable.

crumblingstatue:
Alright, something strange is going on.

I started a clean build, and it indeed fails to compile, even on nightly.

crumblingstatue:
Ah, I think I know what's going on. `bitflags` had a breaking change, and our dependency specification in Cargo.toml wasn't strict enough, allowing rust-sfml to use this new, backwards-incompatible version of it.

Thanks for the fix!

Rafagd:
Just for the record, I was using: 
$ cargo version
cargo 0.5.0-nightly (833b947 2015-09-13)
$ rustc --version
rustc 1.4.0 (8ab8581f6 2015-10-27)

PR #116, opened 2016-03-29
Added antialiasing, and a new pong example using AA

mrkgnao:
Here are two pictures, with a pair of modified Pong examples with a ball radius of `30.` and antialiasing levels `0` and `8` respectively.

Without AA

With 8xAA

I have a low-res monitor, but I think the difference is clearly visible.

### DIFF
--- a/src/window/context_settings.rs
+++ b/src/window/context_settings.rs
@@ -42,35 +42,26 @@ pub const CONTEXT_DEBUG: u32 = 1 << 2;
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Copy)]
 pub struct ContextSettings(pub ffi::sfContextSettings);
 
-/// Builder struct for `ContextSettings`
-pub struct ContextSettingsBuilder(pub ContextSettings);
-
-impl ContextSettingsBuilder {
-    /// Creates a new ContextSettingsBuilder.
-    pub fn new() -> ContextSettingsBuilder {
-        ContextSettingsBuilder(ContextSettings(ffi::sfContextSettings {
+impl ContextSettings {
+    /// Creates a new, default ContextSettings.
+    pub fn new() -> ContextSettings {
+        ContextSettings(ffi::sfContextSettings {
             depth_bits: 0,
             stencil_bits: 0,
             antialiasing_level: 0,
             major_version: 2,
             minor_version: 0,
             attribute_flags: CONTEXT_DEFAULT,
-        }))
+        })
     }
 
     /// Modifies `self` to use some specified level of antialiasing.
-    pub fn antialiasing_level(&mut self, level: u32) -> ContextSettingsBuilder {
-        let ContextSettingsBuilder(ContextSettings(ctx)) = *self;
-        return ContextSettingsBuilder(ContextSettings(ffi::sfContextSettings {
+    pub fn antialiasing_level(&mut self, level: u32) -> ContextSettings {
+        let ContextSettings(ctx) = *self;
+        ContextSettings(ffi::sfContextSettings {
             antialiasing_level: level,
             ..ctx
-        }));
-    }
-
-    /// "Finalizes" the `ContextSettings`.
-    pub fn build(&self) -> ContextSettings {
-        let ContextSettingsBuilder(ctx) = *self;
-        ctx
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


crumblingstatue:
I think it would be better if we added optional antialiasing (through e.g. a command line flag) to the pong example, rather than maintain a copy of the pong example with antialiasing.

src/window/context_settings.rs
```diff
@@ -42,6 +41,21 @@ pub const CONTEXT_DEBUG: u32 = 1 << 2;
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Copy)]
 pub struct ContextSettings(pub ffi::sfContextSettings);
 
+impl ContextSettings {
+    /// Creates a new ContextSettings with all parameters 
+    /// default, except `antialiasing_level`
+    pub fn new_with_antialiasing_level(aa_level: u32) -> ContextSettings {
```
crumblingstatue: I don't think adding a `new_with_foo` function for every property we want to set is a composable solution.

We could instead support the [builder pattern](https://doc.rust-lang.org/book/method-syntax.html#builder-pattern), so the user could do e.g. `let ctx = ContextSettings::default().antialiasing(8);`.

mrkgnao: I added a TODO saying exactly that in my local copy of the repo after
sending the PR. Why don't you let me do it?

I'll submit a new, updated PR today then.

Thanks for the feedback.

mrkgnao:
@crumblingstatue both the builder pattern (with docs) and the command-line argument to Pong are included and working.

crumblingstatue:
The diff shows 11 changed files. Did you really need to change 11 files to implement this?
There is a lot of formatting noise, which makes it hard to find the actual changes you did.

Can you please turn off the auto-reformatting feature of your editor for this pull request?

I eventually plan on running rustfmt on the entire codebase, but I ran into some problems the last time I tried, so I didn't do it yet.

mrkgnao:
@crumblingstatue There you go!

mrkgnao:
@crumblingstatue there you go!

crumblingstatue:
I was thinking of implementing the builder pattern on `ContextSettings` itself instead of using a separate builder, since the two structs are identical, and the only thing the builder methods do is set the various fields. That way, we don't even need a finalizer. The API would look cleaner this way in my opinion.

Here is what it would look like:

```Rust
impl ContextSettings {
    pub fn antialiasing(&mut self, level: u32) -> &mut ContextSettings {
        self.0.antialiasing_level = level;
        self
    }
}

```

And this is how it would be used:
```Rust
RenderWindow::new(VideoMode::new_init(game_width, game_height, 32),
                                       "SFML Pong",
                                       window_style::CLOSE,
                                       ContextSettings::default().antialiasing(aa_level))
```

What do you think?

crumblingstatue:
On a side note, is your system clock running well? Your commits seem to be always in the future.
Github always puts them after the comments, and says `mrkgnao added some commits just now`.

mrkgnao:
Re the builder pattern, I thought so too, but I didn't see anyone do it
that way so I assumed the separate struct thing was something "canonical".
I'll rewrite it that way, then. It'll certainly be cleaner, as you say.

Also, yes, my dual-boot setup has been causing some clock problems for a
while.

mrkgnao:
@crumblingstatue Done. What do you say?

src/window/context_settings.rs
```diff
@@ -42,6 +42,29 @@ pub const CONTEXT_DEBUG: u32 = 1 << 2;
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Copy)]
 pub struct ContextSettings(pub ffi::sfContextSettings);
 
+impl ContextSettings {
+    /// Creates a new, default ContextSettings.
+    pub fn new() -> ContextSettings {
```
crumblingstatue: I'm okay with adding a new() method in addition to the `Default` implementation for consistency, but don't duplicate the initialization code. Use `Default::default()` here instead of copypasting the code.

crumblingstatue:
> My method signature looks a bit different from yours, since it returns a ContextSettings instead of an &mut ContextSettings. Will that be a problem?

Why do you think returning a ContextSettings instead of a &mut ContextSettings is a better solution?

crumblingstatue:
@mrkgnao Are you still interested in completing this pull request?
If not, I could merge it as-is, then implement the desired changes myself on top of it.

mrkgnao:
I'm somewhat busy right now, unfortunately. It would be great if you could
merge it as it is right now.

crumblingstatue:
Thanks for your contribution!

PR #117, opened 2016-04-16
Update required SFML version to latest one

zakkor:


crumblingstatue:
That specific patch version isn't really required, is it?
It should work with any 2.3.x.

zakkor:
Maybe it's on my side, but unless I follow what the travis.ci build script does and compile both CSFML(2.3) and SFML(2.3.2) from source, I get a run time crash when trying to run a simple rsfml "hello world".

crumblingstatue:
> Maybe it's on my side, but unless I follow what the travis.ci build script does and compile both CSFML(2.3) and SFML(2.3.2) from source, I get a run time crash when trying to run a simple rsfml "hello world".

That sounds more like an ABI incompatibility or a bug than a hard requirement on `2.3.2` vs `2.3.x`.

PR #118, opened 2016-04-30
Add a method to retrieve an image's memory buffer.

sebcrozet:
This is useful to save the image using another encoder than sfml's.


sebcrozet:
Thanks for merging! Could the new version be published to crates.io please?

PR #120, opened 2016-05-13
Take `[f32; 9]` instead of 9 `f32` arguments for Transform::new

crumblingstatue:
Instead of taking 9 arguments of the same type, we can take an array.
This looks neater in the documentation, and also satisfies clippy's
`too_many_arguments` lint.


crumblingstatue:
I benchmarked it, the array version has the same performance as the 9 arguments version.

crumblingstatue:
cc @jeremyletang

crumblingstatue:
I'll take the silence as a yes, and go ahead and merge this.

PR #125, opened 2016-05-28
Add bare implementation of custom sound streams

crumblingstatue:
In the original API, this is done by deriving `sf::SoundStream`.

For now, only the minimum required for the included example is implemented.

Oh yeah, this would also bump our **Rust** requirement to **1.9.0**, because  of `panic::catch_unwind`, which is required to make sure panic triggered in user code doesn't unwind across FFI boundaries.

### DIFF
--- a/examples/custom-sound-stream.rs
+++ b/examples/custom-sound-stream.rs
@@ -39,8 +39,8 @@ impl BitMelody {
 }
 
 fn main() {
-    let stream = BitMelody::new();
-    let mut player = SoundStreamPlayer::new(stream, 1, 44100);
+    let mut stream = BitMelody::new();
+    let mut player = SoundStreamPlayer::new(&mut stream, 1, 44100);
     player.play();
     loop {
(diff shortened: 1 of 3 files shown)


crumblingstatue:
CC @jeremyletang, especially regarding Rust 1.9.0 requirement.

I'm also not 100% sure there aren't any soundness issues.

crumblingstatue:
My motivation for implementing this is that I ultimately want to write a [Module](https://en.wikipedia.org/wiki/Module_file) player addon for rust-sfml, since SFML doesn't support them by default. This would be the first "extension" crate for rust-sfml AFAIK.

crumblingstatue:
Here is a crate providing a custom `SoundStream` implementation that plays [module] files: https://github.com/crumblingstatue/rust-sfml-modstream

crumblingstatue:
I think this is ready for merging.

If there are no further comments, I'll merge this tomorrow.

Issue #101, opened 2015-05-23
rc examples are broken


crumblingstatue:
https://travis-ci.org/jeremyletang/rust-sfml/builds/63751034#L472

jeremyletang:
hey, I've recently remove from build the rc part as i planned to refactor some stuff. In the same time @SpaceManiac have propose to change some things on the library, so there will be some breaking change for a time, but this will be fix.
The cargo version stay stable for the users I think.

crumblingstatue:
I commented the rc example out, so the Travis build passes now.
It will need to be uncommented and updated once rc is available again.

PR #126, opened 2016-06-07
Remove non-functional `Rc` versions of resource-using types

crumblingstatue:
These versions haven't been maintained for a long time, possibly because
they are not very amenable to maintenance. They are essentially copy-pasted
versions of the non-`Rc` versions.

Even if we want to support this feature in the future, we should implement
it in a more maintainable way, possibly using generics. If that's not
possible, macros and other forms of code generation are still a more
maintainable solution than the status quo.

cc @jeremyletang


crumblingstatue:
This would also close #101.

crumblingstatue:
I think this is a sensible change, so I'm going to merge it.

PR #133, opened 2016-07-13
Use Unicode functions in Text's string handling (fixes #132)

### LINKED ISSUE #132: `Text` accepts `&str` but doesn't support unicode
Simple test: 
```rust
let text = Text::new_init("ё", &font, 14).unwrap();
window.draw(&text);
```

It shows `ё` as two invalid characters.

See also: “[How to avoid problems with non-ASCII characters?](http://www.sfml-dev.org/tutorials/2.0/graphics-text.php#how-to-avoid-problems-with-non-ascii-characters)”

SpaceManiac:
For #132. The methods accepting `&str` and producing `String` have been updated to use the Unicode-aware C functions, and the unnecessary Unicode variants have been removed.

### DIFF
--- a/src/graphics/text.rs
+++ b/src/graphics/text.rs
@@ -27,7 +27,6 @@
 //! Text is a drawable class that allows to easily
 //! display some text with custom style and color on a render target.
 
-use std::ffi::{CString, CStr};
 use std::str;
 use libc::{c_float, c_uint, size_t};
 
@@ -45,7 +44,7 @@ use csfml_graphics_sys as ffi;
 /// display some text with custom style and color on a render target.
 pub struct Text<'s> {
     text: *mut ffi::sfText,
-    string_length: u32,
+    string_length: usize,
     font: Option<&'s Font>
 }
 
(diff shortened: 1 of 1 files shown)


src/graphics/text.rs
```diff
@@ -107,51 +96,36 @@ impl<'s> Text<'s> {
...
 
-    /// Set the string of a text (from an ANSI string)
+    /// Set the string of a text
     ///
     /// A text's string is empty by default.
     ///
     /// # Arguments
     /// * string - New string
     pub fn set_string(&mut self, string: &str) {
+        let mut utf32: Vec<u32> = string.chars().map(|ch| ch as u32).collect();
```
crumblingstatue: It's unfortunate that we have to do an allocation every time we set the string, but it's better than the currently broken behavior.

Maybe we could also offer separate APIs that make different trade-offs, for example one that takes an ascii string for those who don't need unicode strings, but want to avoid unnecessary allocations. Solving this is not the task of this PR though. This PR *does* fix previously broken functionality, so I'm merging it as-is.

crumblingstatue:
Thanks!

PR #135, opened 2016-07-30
Adds commutativity to operators +, -, *, /

Arandur:
Previously, one could write e.g. `v * 3` but not `3 * v`, where `v` is a Vector2<T> or a Vector3<T>. This commit adds implementations of Add, Sub, Mul, and Div for i32, u32, and f32 upon their respective Vectors, to allow either expression to work.

We can only add implementations of these traits for certain types, due to the restrictions on traits imposed by the language. We still can't, therefore, write `i * v` where `i` is an arbitrary type, even if `i` implements `Mul`. But this commit will at least allow this expression for the six types (Vector[23][uif]) used by the library.

### DIFF
--- a/src/system/vector2.rs
+++ b/src/system/vector2.rs
@@ -1,31 +1,30 @@
-/*
-* Rust-SFML - Copyright (c) 2013 Letang Jeremy.
-*
-* The original software, SFML library, is provided by Laurent Gomila.
-*
-* This software is provided 'as-is', without any express or implied warranty.
-* In no event will the authors be held liable for any damages arising from
-* the use of this software.
-*
-* Permission is granted to anyone to use this software for any purpose,
-* including commercial applications, and to alter it and redistribute it
-* freely, subject to the following restrictions:
-*
-* 1. The origin of this software must not be misrepresented; you must not claim
-*    that you wrote the original software. If you use this software in a product,
-*    an acknowledgment in the product documentation would be appreciated but is
-*    not required.
-*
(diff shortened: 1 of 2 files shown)


src/system/vector2.rs
```diff
@@ -1,31 +1,30 @@
...
+// 1. The origin of this software must not be misrepresented; you must not claim
+//    that you wrote the original software. If you use this software in a product,
+//    an acknowledgment in the product documentation would be appreciated but is
+//    not required.
+//
+// 2. Altered source versions must be plainly marked as such, and must not be
+//    misrepresented as being the original software.
+//
+// 3. This notice may not be removed or altered from any source distribution.
+//
```
crumblingstatue: Please try not to format unrelated parts of the code with rustfmt. It makes it harder to review the actual changes.

src/system/vector3.rs
```diff
@@ -154,6 +141,7 @@ impl<T: Div> Div for Vector3<T> {
     }
 }
 
+
```
crumblingstatue: Looks like an unneeded extra newline here

crumblingstatue:
These additions look okay to me. It's unfortunate that we can't blanket impl the traits, but there's nothing we can do about that right now. This doesn't seem too much of a maintenance burden, and symmetry is always nice.

If no one objects about this change, I'll merge it tomorrow.

src/system/vector2.rs
Arandur: Oh, yeah, my bad. I'll ensure that doesn't happen again.

PR #137, opened 2016-10-03
Using target_env and CSFML_HOME in build scripts.

massung:
Uses cfg!(target_env) in system build scripts to get the name
of the library to link correct:

target_env(msvc) -> csfml-*
target_env(gnu) -> libcsfml-*

Also added an option CSFML_HOME environment variable check so
the libraries can be located somewhere other than the root
directory of the project using SFML.


massung:
I don't mind doing it, but haven't used common build scripts before with Cargo (not sure how to "include" it). I'll look, but if you have an example to point to, I'll make it so.

crumblingstatue:
> I don't mind doing it, but haven't used common build scripts before with Cargo (not sure how to "include" it). I'll look, but if you have an example to point to, I'll make it so.

The easiest way is to use the `#[path="..."]` attribute on a mod declaration. This way, you can, for example, put `#[path="../build_common.rs"] mod build_common;` in each build script.

massung:
Done. Thanks for the info on how to do it.

build_sys_common.rs
```diff
@@ -0,0 +1,31 @@
...
+// for the libraries, otherwise it defaults to `.`.
+//
+// It assumes that CSFML is simply unzipped - as is - into a folder, and the
+// libraries exist in: `$CSFML_HOME/lib/[msvc|gcc]/*.lib`.
+//
+pub fn lib_path(file: &str) -> String {
+    let lib = if cfg!(target_env = "msvc") {
+        format!("csfml-{}", file)
+    } else {
+        format!("libcsfml-{}", file)
```
crumblingstatue: Is this how the lib files are named on windows with the GNU toolchain?

Because on *nix, they are named simply `csfml-{}`, and `libcsfml-{}` is invalid, causing linking to fail on my Linux system.

massung: I don't have Linux to test under. So, it might need another cfg! check, sadly. But, if just download CSFML from the main page and extract the files, in the lib/gcc folder is "libcsfml-audio.a", "libcsfml-graphics.a", etc.

massung: Downloading the OS X version also has them named `libcsfml-*.dylib`. I guess two questions:

1. Does rust auto-add the "lib" prefix? If so, then the MSVC check can go away and just leave the CSFML_HOME there (assuming you want it).
2. Is there something you'd like me to do to the code w/o being able to test it?

crumblingstatue: Oh, yeah, I realize what's up now.

Under the GNU toolchain, library files are always prefixed with `lib`, but the `lib` prefix is not included when supplying the library name to the linker. It's just how the GNU linker works I guess.

So for example, the library is named `libfoo.a`, but the library name you pass to the linker is just `foo`.

This is how it works on Unix, but I believe this is how it works on Windows under the GNU toolchain as well. Did you test this on Windows with the GNU toolchain? It should fail to link with `libcsfml-*`, but should succeed with `csfml-*`.

massung: Actually, downloading all the packages, looks like the windows version is the only one that separates out msvc/libs and gcc/libs. So, need to fix that regardless.

massung: Okay, I'll put together a change that gets rid of the "lib" prefix, and also fixes the directory issue per-platform. But, (and sorry), I won't be able to test it for you outside of windows. :(

crumblingstatue:
> Does rust auto-add the "lib" prefix? If so, then the MSVC check can go away and just leave the CSFML_HOME there (assuming you want it).

The lib prefix is not needed, see my above comment.



> Is there something you'd like me to do to the code w/o being able to test it?

Just removing the lib prefix should be fine. If you can afford to, however, it would be nice if you could test this on windows with the GNU toolchain.

massung:
Rust 1.12.0 windows MSVC and GCC both build fine. Changed the workings of the shared build code, because GCC linker wanted link paths instead of a full path to a lib. Should work with *nix and OS X given the downloads on the SFML website. Works fine if no CSFML_HOME is set and the libs are in the local project folder, too (same as w/o the PR).

crumblingstatue:

Issue #149, opened 2016-10-22
Consider supporting dynamic dispatch for Drawable::draw


crumblingstatue:
We might want to consider supporting this, because the C++ SFML API also supports dynamic dispatch of `RenderTarget`.

crumblingstatue:
Closed by #154

PR #154, opened 2016-12-05
Make RenderTarget and Drawable traits object-safe

sleeparrow:
Hi! I think object-safe traits are preferable, but my Rust is a little rusty. Let me know if there's any cleanup needed or if you don't think this change is a good idea. :)


sleeparrow:
cc #108 and #149

crumblingstatue:
Since the C++ API does the same thing, I'm accepting this.

I don't think this should have a significant effect on drawing performance.
I did some rudimentary testing, and this change didn't seem to have a negative effect on performance.

Issue #138, opened 2016-10-08
Better documentation and support for finding native libs on windows


crumblingstatue:
1. [ ] Determine where the SFML libraries should be installed for each of the GNU and MSVC toolchains
2. [ ] Document it
3. [ ] Add support in the build scripts for detecting and using the libraries installed in these locations.

Additional things that can be done
- [ ] Document #137

crumblingstatue:
Mentioning #165 for reference.

TheJare:
I just got a rust-sfml window running on Windows with MSVC. It was painful because the documentation for building sfml/csfml with CMake is pretty much nonexistent to begin with.

Apparently, CSFML needs to be built as dynamic libraries, but use SFML as static libraries. So this is what I did, ending up with the following:

- sfml-XXX-s.lib files
- csfml-XXX.lib files
- csfml-XXX-2.dll files

I had to rename the sfml-XXX-s.lib to remove the "-s" suffix from the name, because this crate wants the clean name. I don't know if this suffix is something peculiar to the CMake build of SFML or what.

But, thinking about it more... if CSFML are DLLs containing a statically linked SFML, and the rust-sfml crate only interfaces with CSFML, then rust-sfml should not need to link with SFML at all, since it does not use it. Is this correct?

crumblingstatue:
> But, thinking about it more... if CSFML are DLLs containing a statically linked SFML, and the rust-sfml crate only interfaces with CSFML, then rust-sfml should not need to link with SFML at all, since it does not use it. Is this correct?

This might be true on windows, you could test it if you have the time. It isn't true on linux, where csfml links to sfml dynamically.

In the long term, I'm planning to skip CSFML entirely, and use bindgen to bind directly to SFML, but I don't know yet what will come of that.

lernrust:
Hi. I have 2 problems :(

SFML 2.4.0 
CSFML 2.4.0
32bit

E:\Users\Admin\projects\rust-sfml> cargo build --verbose
       Fresh bitflags v1.0.1
       Fresh sfml-build v0.2.0 (file:///E:/Users/Admin/projects/rust-sfml/ffi/sfml-build)
       Fresh csfml-system-sys v0.4.0 (file:///E:/Users/Admin/projects/rust-sfml/ffi/csfml-system-sys)
       Fresh csfml-window-sys v0.4.0 (file:///E:/Users/Admin/projects/rust-sfml/ffi/csfml-window-sys)
[...]

PR #157, opened 2017-03-03
Finding CSFML and SFML dependencies on Windows

bookdude13:
When building on Windows rust-sfml wasn't able to find my CSFML and SFML libraries. Found out that the paths in the build scripts were using forward slashes, not back slashes. Updated them, and also had them look for SFML libraries at SFML_HOME, not just CSFML libraries at CSFML_HOME.

This addresses #138 (I think). I bumped the versions; can change back if you want.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "sfml"
 description = "Rust binding for sfml"
-version = "0.11.1"
+version = "0.11.3"
 authors = ["Jeremy Letang <letang.jeremy@gmail.com>"]
 
 documentation = "https://docs.rs/sfml"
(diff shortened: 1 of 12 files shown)


crumblingstatue:
> I bumped the versions; can change back if you want.

Yeah, bumping versions should generally only be done before releasing a new version to crates.io.

I'll review this soon, in the meantime, you can revert the version bumps.

bookdude13:
Version bumps reverted. 

csfml-window-sys lists itself as version 0.1.1 in its Cargo.toml, but it was originally listed as 0.1.0 in rust-sfml's Cargo.toml. Should this stay bumped up?

Along the same lines, rust-sfml shows up as version 0.11.2 in crates.io, but the Cargo.toml was set to 0.11.1. Same question, should this stay bumped up to be accurate, or should they still be bumped down?

crumblingstatue:
This pull request should focus on finding CSFML on windows, and leave everything else alone, including dependency versions. They can be sorted out before a new crates.io release.

bookdude13:
Understood. Those versions now reverted.

crumblingstatue:
Thanks for your contribution!

nabijaczleweli:
> forward slashes on windows work just as well as backward slashes

They do and are not as ugly. Also, I'd recommend using `r"thing\other\thing"` to escaping for clarity.

crumblingstatue:
@bookdude13: Maybe you could check if using forward slashes works. Maybe something else was the problem. It's not that big of a deal to have a few extra branches for windows paths though.

bookdude13:
Will double check, but pretty sure it failed with forward slashes, then worked with back slashes. The slashes are combined with whatever is in CSFML_PATH, so might be an issue with mixed forward and back slashes.

bookdude13:
Double checked. @nabijaczleweli and @crumblingstatue you were both right, the forward/back slashes weren't the issue, it was putting stuff in the right place. This pull request didn't really fix finding the dependencies, but it made it so SFML libraries can be located in a different location than CSFML libraries, which was part of my issue.

If you want I can set the paths back to forward slashes; I agree that it is cleaner.

crumblingstatue:
> If you want I can set the paths back to forward slashes; I agree that it is cleaner.

Up to you, you can submit a pull request if you'd like.

PR #158, opened 2017-03-05
Cleaned up build script paths

bookdude13:
Made back slashes forward slashes so the paths look cleaner. Also simplified the SFML logic, since both Windows and non-WIndows systems have the SFML libraries in the lib/ subdirectory. This is an extension of #157.

On a side note, if I end up doing more pull requests, should I make an issue to match the pull request? Or should I just do the pull request? Don't know if you have a process or not.


crumblingstatue:
> On a side note, if I end up doing more pull requests, should I make an issue to match the pull request? Or should I just do the pull request? Don't know if you have a process or not.

Simply a pull request is enough if the motivation is clear, or you explain it in the pull request.

PR #159, opened 2017-03-08
Split the different SFML modules out into their own crates

crumblingstatue:
# Advantages

1. The user only needs to compile and link the module that they actually use.

     For example, if the user wants to create a command line audio application, they only need to depend on `sfml-system` and `sfml-audio`. They don't need to link against the other libraries.

    Another example is the `sfml-network` crate. It is almost _never_ used, and barely maintained. Rust already has better tools for this, even in the standard library.

2. We can release new versions of the different modules independently. If one module is ready for a new release, it doesn't have to wait on the others.

3. Better potential integration with other crates.

    For example, `sfml-window` could be a [Piston](https://github.com/PistonDevelopers/piston) window backend, without bringing in a ton of useless dependencies.

# Disadvantages

1. The user has to declare multiple dependencies for multiple modules.

     I don't consider this to be too much of a burden. A possible solution would be to keep the current `sfml` crate as a thin layer that reexports the APIs of the different crates.

2. Minor API leakage of some internals. For example, `SfBoolExt` is now public. I made it `doc(hidden)` though, since it's of no use to the user. I don't think anything leaked that could cause unsoundness (hopefully).

3. The examples are now scattered among the different crates instead of being in a central place. We could possibly create an `sfml-examples` repository, or something similar to remedy this.

I personally believe this is the right way to do it, since the C++ SFML library also allows only linking against the modules that you actually use.

# Alternative

An alternative would be to use Cargo features, [like sdl2 does](https://github.com/AngryLawyer/rust-sdl2/blob/master/Cargo.toml#L41-L47). This would also give us advantage 1, and not have disadvantages 1, 2, and 3. This is actually a good compromise now that I think about it.

---

I call out for people who have an interest in rust-sfml to review this, and give their opinion.

Owner: @jeremyletang 

(reasonably recent) Past contributors:
- @bookdude13 
- @sleeparrow 
- @juliendehos 
- @massung 
- @Arandur 
- @SpaceManiac 
- @sebcrozet 
- @mrkgnao 
- @brandonson 

Unresolved questions:
- Who should publish the new crates? I am willing to publish them myself, but perhaps @jeremyletang would like to retain ownership of the project on crates.io.


massung:
A potential disadvantage not mentioned:

* Each crate can be independently updated. This could potentially run into issues later with one (or more) crates outpacing the others and leading to a mismatch of features/functionality. Since `sfml-graphics` depends on `sfml-window`, a change to `sfml-window` may pass its own internal tests, but break the latest version of `sfml-graphics`.

That's not a deal breaker by any means since `sfml-graphics` could (and should) denote the version of `sfml-window` that it works with in its TOML file, but it does make things slightly more annoying for users. I mostly bring it up because it's something that [sdl2](https://crates.io/crates/sdl2) users have to deal with right now when trying to also use [sd2_image](https://crates.io/crates/sdl2_image), etc. Currently, SFML has the advantage of being a one-stop-shop.

It might be nice to have a `sfml-all` module that pulled everything at known good versions together. This way, users will still have a single place to go to for the majority of use-cases.

My 2 cents.

sleeparrow:
> Better potential integration with other crates.
> For example, sfml-window could be a Piston window backend, without bringing in a ton of useless dependencies.

I have limited experience with Piston, but I found the modularity of the library made everything a _huge_ pain.

I'm not saying I'm against what's being suggested here; I'm not even saying I'm against making rust-sfml modular, but please actively test the library.

I tend to prefer convention over customization. Keep an eye on the ease of use of the library, as you can easily complicate things by being too "flexible".

crumblingstatue:
@massung: `sdl2` also had the disadvantage of `sdl2` and `sdl2_image`, `sdl2_mixer`, etc. had different repositories and different maintainers, so they weren't kept in sync. This is not the case with `sfml`.

I've recently noticed that the various sdl2 addons were merged into the main repository, and it's [using features](https://github.com/AngryLawyer/rust-sdl2/blob/master/Cargo.toml#L41-L47) for enabling the addons wanted by the user. This could be an alternative to be considered in our case too.

crumblingstatue:
I'm considering using features instead of this solution. If in the future a crate split would be advantageous, it can still be done, but using features would solve the most prominent use case, which is to only link against the libraries you are using.

I'll open an alternative pull request that uses features, and see how it works out.

jeremyletang:
I'm not taking part of the development anymore, so do as you like / you think is best, but for what it worth, I think that what make the c++ version of the sfml really great is, even if you have to link to multiple library, it's kind of a "all in one" library, where you get anything you need to write a game from the library easily from one place.

For sure the rust version may not want to link against all the library if you use only one of them, and for this the features may be the best way to implement it.

By default you may want to link against all the sfml library, maybe just let the user explicitly pick the one he want if he do not need everything.

crumblingstatue:
I'm closing this in favor of #160, as that is a much less controversial change.

PR #160, opened 2017-03-09
Make the different modules optional using Cargo features

crumblingstatue:
This enables users to only link to the SFML libraries that they are using.

Alternative to #159.

By default, all modules are enabled, but the user can opt-out with `default-features = false`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,12 +9,17 @@ repository = "https://github.com/jeremyletang/rust-sfml"
 license = "zlib-acknowledgement"
 
 [features]
+default = ["graphics", "audio", "network"]
 # Used for conditonally setting clippy lint levels.
 # Run cargo clippy with `cargo clippy --features=clippy`.
 clippy = []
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -9,12 +9,17 @@ repository = "https://github.com/jeremyletang/rust-sfml"
 license = "zlib-acknowledgement"
 
 [features]
+default = ["graphics", "audio", "network"]
```
crumblingstatue: Note that `window` is implied by `graphics`, so it doesn't have to be explicitly included here.

tests/network.rs
```diff
@@ -1,8 +0,0 @@
-extern crate sfml;
```
crumblingstatue: This never really made sense as an integration test anyway, So I moved it into a unit test inside the network module.

massung:
I definitely like this approach better than #159.

crumblingstatue:
Alright, this seems sensible to merge, so I'm doing it.

PR #169, opened 2017-10-04
Add get_samples method to SoundBufferRef

### LINKED ISSUE #168: Get audio samples from a buffer
It is nice to have a simple audio buffer. But what good is it if I cant access the samples in it? The documentation of the `sample_count` function specifies that there should be a `get_samples` function... but it sure is not present on the buffer itself.
screenshot_20171004_193906

petoknm:
Fixes #168

### DIFF
--- a/src/audio/sound_buffer.rs
+++ b/src/audio/sound_buffer.rs
@@ -101,8 +101,9 @@ impl SoundBufferRef {
     }
 
     /// Get the samples stored in the buffer
-    pub fn get_samples(&self) -> &[i16] {
-        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), self.sample_count() as usize) }
+    pub fn samples(&self) -> &[i16] {
+        let len = self.sample_count().max(0).min(usize::max_value() as i64) as usize;
+        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), len) }
     }
 
     /// Get the number of channels used by a sound buffer
--- a/src/audio/sound_buffer.rs
+++ b/src/audio/sound_buffer.rs
@@ -102,8 +102,11 @@ impl SoundBufferRef {
 
     /// Get the samples stored in the buffer
     pub fn samples(&self) -> &[i16] {
-        let len = self.sample_count().max(0).min(usize::max_value() as i64) as usize;
-        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), len) }
+        let len = self.sample_count();
+        if len > usize::max_value() as u64 {
+            panic!("Sample count {} too big to fit into usize", len);
+        }
+        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), len as usize) }
     }
 
     /// Get the number of channels used by a sound buffer
--- a/src/audio/sound_buffer.rs
+++ b/src/audio/sound_buffer.rs
@@ -5,6 +5,7 @@ use std::borrow::{Borrow, ToOwned};
 use std::ffi::CString;
 use std::io::{Read, Seek};
 use std::ops::Deref;
+use std::slice;
 use system::Time;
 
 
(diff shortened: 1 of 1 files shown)


crumblingstatue:
I'll review this later today, sorry for the delay

src/audio/sound_buffer.rs
```diff
@@ -99,6 +100,11 @@ impl SoundBufferRef {
         unsafe { ffi::sfSoundBuffer_getSampleCount(self.raw()) as i64 }
     }
 
+    /// Get the samples stored in the buffer
+    pub fn get_samples(&self) -> &[i16] {
```
crumblingstatue: The current naming convention is to omit the `get` part where it makes sense. That would make this function named just `samples`.

src/audio/sound_buffer.rs
```diff
@@ -99,6 +100,11 @@ impl SoundBufferRef {
...
+        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), self.sample_count() as usize) }
```
crumblingstatue: This can potentially cause undefined behavior on platforms where `usize` is 32 bits and `sample_count()` returns a value greater than u32::MAX. Perhaps we should check for this condition.

crumblingstatue:
If you'd like, I can merge this pull request as-is, and do the requested changes myself.

src/audio/sound_buffer.rs
petoknm: I would also like that better... I just used get_samples because the sample_count documentation talked about it.

petoknm:
I'll make the changes. I was also wondering if I should put the function in `SoundBuffer` or `SoundBufferRef`... Im kinda new to rust... when i do sound_buffer.samples() does it first not find it and try to dereference it or how does it find a method on the `SoundBufferRef`

src/audio/sound_buffer.rs
petoknm: So if the `sample_count` is bigger than `u32::max_value` what should happen? should it return and `Option<&[i16]>`?

petoknm: Or should I just cap it to `u32::max_value()` length

crumblingstatue:
> I was also wondering if I should put the function in SoundBuffer or SoundBufferRef

`SoundBufferRef` is the appropriate type here.

> when i do sound_buffer.samples() does it first not find it and try to dereference it

That's right. `SoundBuffer` implements [`Deref<Target=SoundBufferRef>`](https://docs.rs/sfml/0.13.2/sfml/audio/struct.SoundBuffer.html#deref-methods).

src/audio/sound_buffer.rs
crumblingstatue: Since this is such an unlikely scenario, I think it's appropriate to simply panic here.

crumblingstatue: If it was stable, we could use the [TryFrom](https://doc.rust-lang.org/nightly/std/convert/trait.TryFrom.html) API, which would be easier than manually implementing this.

petoknm:
I wasn't sure about how to clamp the length properly. I used min and max to clamp it as i64 and then cast it to usize. I'm hope there is a better way to do it but im just not aware of it.

src/audio/sound_buffer.rs
```diff
@@ -101,8 +101,9 @@ impl SoundBufferRef {
     }
 
     /// Get the samples stored in the buffer
-    pub fn get_samples(&self) -> &[i16] {
-        unsafe { slice::from_raw_parts(ffi::sfSoundBuffer_getSamples(self.raw()), self.sample_count() as usize) }
+    pub fn samples(&self) -> &[i16] {
+        let len = self.sample_count().max(0).min(usize::max_value() as i64) as usize;
```
crumblingstatue: Could you rebase this on top of master? I added a commit that fixes the signature of sample_count to be `u64` instead of the incorrect `i64`.

I also think it would be more appropriate to panic than silently clamp the value.

crumblingstatue:

petoknm:
I'm glad I could help 😄 . When approximately can I expect these changes to be released?

crumblingstatue:
> When approximately can I expect these changes to be released?

I can release a new version now. Hang on.

petoknm:
Awesome! Thanks

crumblingstatue:
[0.14.0 published](https://crates.io/crates/sfml/0.14.0).

PR #171, opened 2017-10-07
[WIP] Add SoundRecorder

### LINKED ISSUE #170: SoundRecorder functionality
In C++ was able to create my own audio stream handler [like this](https://github.com/petoknm/fft-analyzer/blob/master/listener.h). I cannot find a way how to do this with the rust bindings. I need to be able to process data as it comes in not in bulk like with `SoundBufferRecorder`.  Is there a way to extend the bindings to be able to do this?

petoknm:
Fixes #170. This is work in progress PR for adding `SoundRecorder` functionality to rust-sfml. I also created an example on how the the functionality could be used. I welcome any suggestions on how to improve the API. I still need to add documentation and work a bit more on the unsafe functions. Now the compiler is optimizing the closures during monomorphisation so much that I don't even need the `void*` because the function call gets inlined directly. But syntactically I still need an instance of the type of the closure to call it (I think). If there is a way to call a closure just by knowing it's type please let me know 😃.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,10 @@ name = "custom-shape"
 required-features = ["graphics"]
 
 [[example]]
+name = "custom-sound-recorder"
+required-features = ["audio"]
+
+[[example]]
 name = "custom-sound-stream"
 required-features = ["audio"]
 
--- a/examples/custom-sound-recorder.rs
+++ b/examples/custom-sound-recorder.rs
@@ -1,0 +1,26 @@
+extern crate sfml;
+
+use std::{thread, time};
+use sfml::audio::SoundRecorder;
+
+fn main(){
+
+    let sr = SoundRecorder::new(
+        || {
+            println!("Recording started!");
+            true
+        },
+        |_| {
+            println!("Samples received!");
+            true
+        },
+        || {
+            println!("Recording stopped!");
+        }
+    );
+
+    sr.start(44100);
+    thread::sleep(time::Duration::from_secs(3));
+    sr.stop();
+
+}
--- a/examples/custom-sound-recorder.rs
+++ b/examples/custom-sound-recorder.rs
@@ -5,7 +5,7 @@ use sfml::audio::SoundRecorder;
 
 fn main(){
 
-    let sr = SoundRecorder::new(
+    let mut sr = SoundRecorder::new(
         || {
             println!("Recording started!");
             true
--- a/examples/custom-sound-recorder.rs
+++ b/examples/custom-sound-recorder.rs
@@ -10,8 +10,8 @@ fn main(){
             println!("Recording started!");
             true
         },
-        |_| {
-            println!("Samples received!");
+        |samples| {
+            println!("Samples received! {}", samples.len());
             true
         },
         || {
@@ -19,6 +19,7 @@ fn main(){
         }
     );
 
+    sr.set_processing_interval(time::Duration::from_millis(10));
     sr.start(44100);
     thread::sleep(time::Duration::from_secs(3));
     sr.stop();
--- a/src/audio/mod.rs
+++ b/src/audio/mod.rs
@@ -6,10 +6,14 @@ extern crate csfml_audio_sys;
 pub use self::music::Music;
 pub use self::sound::Sound;
 pub use self::sound_buffer::{SoundBuffer, SoundBufferRef};
+pub use self::sound_recorder::SoundRecorder;
 pub use self::sound_buffer_recorder::SoundBufferRecorder;
 pub use self::sound_source::SoundSource;
 pub use self::sound_status::SoundStatus;
 pub use self::sound_stream::{SoundStream, SoundStreamPlayer};
+
+#[derive(Debug, Clone, Copy)]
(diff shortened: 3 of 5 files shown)


petoknm:
I also moved the `SetDeviceError` up to the audio module because it is used by both `SoundRecorder` and `SoundBufferRecorder`

petoknm:
I see that for the `SoundBufferRecorder` you borrow mutably self for in method `set_device`. Is is just for the sake of the API? because the code seems like it would be fine with an immutable reference.

crumblingstatue:
> see that for the SoundBufferRecorder you borrow mutably self for in method set_device. Is it just for the sake of the API? because the code seems like it would be fine with an immutable reference.

`sfSoundBufferRecorder_setDevice` takes `*mut sfSoundBufferRecorder`, which means it most likely mutates the underlying recorder. Doing this through an immutable reference would be unsound.

petoknm:
but that is not enforced by the compiler... as far as I can see... because the compiler seems to happily compile it with an immutable reference to `SoundRecorder` (in my case) and a field on that object is a `*mut ...` that i pass to the ffi function

crumblingstatue:
> but that is not enforced by the compiler... as far as I can see... because the compiler seems to happily compile it with an immutable reference to SoundRecorder (in my case) and a field on that object is a *mut ... that i pass to the ffi function

We are dealing with raw pointers and unsafe code here. For borrowed references, Rust protects you from making mistakes, but not for raw pointers. Maintainers of unsafe code need to be extra careful, and make sure they know what they are doing, because the compiler is no longer holding their hands.

petoknm:
Ok. Got it. So im gonna make the method take self as a mutable reference. Thanks. What do you think about the API so far?

petoknm:
And also there are some other functions shared by `SoundRecorder` and `SoundBufferRecorder`. Namely `is_available`, `default_device`, `available_devices`. Should they move to a separate file in the audio module?

crumblingstatue:
> What do you think about the API so far?

I'm in the process of evaluating it. My main concern so far is the unfriendliness towards reusable types that keep track of state. How would you implement `MyCustomRecorder` that updates `self` when new samples are received, for example?

> And also there are some other functions shared by SoundRecorder and SoundBufferRecorder. Namely is_available, default_device, available_devices. Should they move to a separate file in the audio module?

If you have a good idea on how to structure it in a DRY manner, then go ahead. I don't right now.

petoknm:
I really have no idea how to make it easily extendable and I realize that it is an important feature of the API. The only way I can think of right now would be to make this `SoundRecorder` foldable. And then you could give it the initial value and fold all samples through. I think that implementing an Iterator trait could solve some of these issues.

crumblingstatue:
If you take a look at the existing `SoundStream` API, you can see that [it allows what I described](https://github.com/jeremyletang/rust-sfml/blob/54b984e5d7c1f5e4a558571172d2f1ec25b8feaa/examples/custom-sound-stream.rs).

I believe the same could be done for `SoundRecorder`. The only annoying part about the API is having to use a `SoundStreamPlayer` to play it, but I'm thinking about an alternate solution.

petoknm:
Conceptually it is the same as a sound stream with the only difference being that it is push based instead of pull based. I don't think it would be a good idea to convert it to pull based interface because we would then have to copy the buffer over on every callback.

crumblingstatue:
Having to use CSFML's callback based interface might severely limit our design choices here.

I'm seriously considering using bindgen to directly interface with the C++ API, so we don't have to go through CSFML.

petoknm:
How would using the C++ SFML API change our API options?

crumblingstatue:
> How would using the C++ SFML API change our API options?

To be honest, I'm not sure. I was thinking that it might allow us to use inheritance in a limited fashion, so we don't have to go through the callback interface of CSFML, but now that I read more about bindgen, it doesn't seem to support cross-language inheritance.

petoknm:
Inheritance is also not part of idiomatic rust as far as I'm aware... Using the C API is the same just that instead of using a vtable of function pointers (virtual methods in C++) it uses `void*` to get to the callback function

crumblingstatue:
> Using the C API is the same just that instead of using a vtable of function pointers (virtual methods in C++) it uses void* to get to the callback function

The main problem is that you can't change the userdata pointer after creating the sfSoundRecorder, so for example, if the data gets moved, you can't change it to point to the new location. This forbids us from creating types that can be moved, and still be used as a `SoundRecorder`.

This problem could be solved by changing CSFML to support this though.

petoknm:
Correct. We wouldn't be able to move them. I'll have to look at how the CSFML works. Not sure if we could change it. I think we could implement moving with just copying the pointer to the CSFML sound recorder and setting the pointer to null on the other side, so that it doesn't drop it.

crumblingstatue:
I made an alternate implementation in #172. This follows the same API that was established with `SoundStream`. Does this suit your use case, or are there problems with it?

petoknm:
Yes that one looks nice and should fit all my needs and needs of others hopefully too.

crumblingstatue:
Closing in favor of #172

PR #175, opened 2017-11-06
Implemented MulAssign and DivAssign between two vectors

tforgione:
Hi.

First, thank you for this great binding of a great library.

I noticed that in the C++ version of SFML, the operators `*=` and `/=` are implemented between vectors, so you can do something like `v1 *= v2` where `v1` and `v2` are vectors (as you can see in [this doc](https://www.sfml-dev.org/documentation/2.4.2/classsf_1_1Vector2.php) for example).

When I tried to do that in rust-sfml, I found out that the traits were not implemented. We can do `v1 = v1 * v2`, but not `v1 *= v2`.

This pull request implements the traits `MulAssign<VectorX<T>>` and `DivAssign<VectorX<T>>` so we are able to do it.

### DIFF
--- a/src/system/vector2.rs
+++ b/src/system/vector2.rs
@@ -151,6 +151,13 @@ impl<T: MulAssign + Copy> MulAssign<T> for Vector2<T> {
     }
 }
 
+impl<T: MulAssign + Copy> MulAssign<Vector2<T>> for Vector2<T> {
+    fn mul_assign(&mut self, rhs: Vector2<T>) {
+        self.x *= rhs.x;
+        self.y *= rhs.y;
+    }
+}
+
 impl<T: Div> Div for Vector2<T> {
     type Output = Vector2<T::Output>;
 
@@ -166,6 +173,13 @@ impl<T: DivAssign + Copy> DivAssign<T> for Vector2<T> {
     fn div_assign(&mut self, rhs: T) {
         self.x /= rhs;
         self.y /= rhs;
+    }
+}
+
+impl<T: DivAssign + Copy> DivAssign<Vector2<T>> for Vector2<T> {
+    fn div_assign(&mut self, rhs: Vector2<T>) {
+        self.x /= rhs.x;
+        self.y /= rhs.y;
     }
 }
 
--- a/src/system/vector3.rs
+++ b/src/system/vector3.rs
@@ -154,6 +154,14 @@ impl<T: MulAssign + Copy> MulAssign<T> for Vector3<T> {
     }
 }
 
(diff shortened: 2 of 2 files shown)


crumblingstatue:
> I noticed that in the C++ version of SFML, the operators *= and /= are implemented between vectors, so you can do something like v1 *= v2 where v1 and v2 are vectors (as you can see in this doc for example).

I don't see any functions documented that are multiplication, and accept `Vector2` on both sides.

Furthermore, the following fails to compile:

```C++
#include <SFML/System.hpp>
#include <iostream>

int main() {
    sf::Vector2f v(2, 4);
    sf::Vector2f v2(4, 8);
    v *= v2;
    std::cout << v.x << ", " << v.y << std::endl;
}

```

Are you sure you can do this in the C++ API?

tforgione:
Oops, you're right. I got confused when I read the docs... I thought I was reading the multiplication assignement operator while it was just the multiplication operator... sorry for that.

I'll create a thread on the SFML forum for this because I feel like it makes sense.

Thanks for the quick answer !

tforgione:
Ok, I think I got why I was confused.

Tell me if I'm wrong but: 
  - in the C++ version of SFML, there is no `operator*` between two vectors. Doing 
    ```cpp
    #include <iostream>
    #include <SFML/System/Vector2.hpp>
    
    int main(int argc, char *argv[])
    {
        sf::Vector2f v(2, 4);
        sf::Vector2f v2(4, 8);
        v = v * v2;
        std::cout << v.x << ", " << v.y << std::endl;
    }
    ```
    will produce a compilation error: (`error: invalid operands to binary expression ('sf::Vector2f' (aka Vector2<float>') and 'sf::Vector2f'`)

  - However, the rust equivalent will work (e.g. because of the [following trait implementation](https://github.com/jeremyletang/rust-sfml/blob/ef928004aa39132f5e08a3606d2a4f7144020c15/src/system/vector2.rs#L136-L145)). There are also some other operators that are implemented, e.g. the addition between a vector and a number.
    ``` rust
    extern crate sfml;
    use sfml::system::Vector2f;
    
    fn main() {
        let v1 = Vector2f::new(2.0, 4.0);
        let v2 = Vector2f::new(1.0, 2.0);
        println!("{:?}", v1 * v2);
        println!("{:?}", v1 + 2.0);
    }
    ```
    will output
    ```
    Vector2 { x: 2, y: 8 }
    Vector2 { x: 4, y: 6 }
    ```

Anyway, thanks for having taken the time to read my comments and to have tried what I didn't. It clarified my knowledge about how operator overloading is done in SFML.

crumblingstatue:
Thanks!

I think I'll remove the operations that SFML doesn't have, so we are consistent with the original API.

PR #185, opened 2018-02-08
Remove unnecessary mut.

datMaffin:
A seemingly unnecessary `mut`.


crumblingstatue:
This does indeed look like an unnecessary `mut`, thanks!

PR #189, opened 2018-06-07
Add ability to create the SFML window from an existing handle

zstewar1:
Add Window::from_handle and RenderWindow::from_handle as unsafe functions.

These would support creating windows from platform-specific window handles, such as HWIND on windows or an XLib Window on Linux. This allows SFML to be used to draw on top of windows allocated by other programs, such as for X screen savers or live desktops.

These functions are marked as unsafe because the caller is responsible for ensuring that the handle is valid. In my manual testing (only on Linux), these would crash with an X error if the window handle is invalid rather than panic at `assert!(!sf_render_win.is_null(), "Failed to create Window");`, so I'm not sure any additional error handling is possible.

### DIFF
--- a/src/graphics/render_window.rs
+++ b/src/graphics/render_window.rs
@@ -58,6 +58,32 @@ impl RenderWindow {
         }
     }
 
+    /// Create a render window from an existing platform-specific window handle
+    ///
+    /// This function creates a render window based on an existing platform
+    /// specific window handle which has been allocated outside of SFML. This is
+    /// only intended to be used in cases where you need to integrate SFML with
+    /// some other windowing library.
+    ///
+    /// This function is unsafe because it is the caller's responsibility to
+    /// ensure that it is called with a valid window handle.
+    ///
+    /// # Arguments
+    /// * handle - The handle to the platform-specific window handle to use for
+    ///            the window.
+    /// * settings - Additional settings for the underlying OpenGL context
+    pub unsafe fn from_handle(
+        handle: ::csfml_window_sys::sfWindowHandle,
+        settings: &ContextSettings,
+    ) -> RenderWindow {
+        let sf_render_win: *mut ffi::sfRenderWindow =
+            ffi::sfRenderWindow_createFromHandle(handle, &settings.raw());
+        assert!(!sf_render_win.is_null(), "Failed to create Window");
+        RenderWindow {
+            render_window: sf_render_win,
+        }
+    }
+
     /// Change a render window's icon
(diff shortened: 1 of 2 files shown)


crumblingstatue:
Sorry for making you wait, I'll review this.

src/graphics/render_window.rs
```diff
@@ -58,6 +58,32 @@ impl RenderWindow {
...
+    ///
+    /// This function is unsafe because it is the caller's responsibility to
+    /// ensure that it is called with a valid window handle.
+    ///
+    /// # Arguments
+    /// * handle - The handle to the platform-specific window handle to use for
+    ///            the window.
+    /// * settings - Additional settings for the underlying OpenGL context
+    pub unsafe fn from_handle(
+        handle: ::csfml_window_sys::sfWindowHandle,
```
crumblingstatue: If possible, I would prefer if we didn't have direct references to CSFML in the public API. The user should only have to care about `sfml`, and not `csfml_*_sys`. For example, if they wanted to store the window handle in a struct, they would have to do `my_handle: ::csfml_window_sys::sfWindowHandle`, and declare `csfml_window_sys` as an extern crate.

How about we create a public `sfml::window::Handle` type, which can then be an alias to the CSFML type. Then, the user can simply do `my_handle: window::Handle`.

zstewar1: Done. Let me know if you want me to put it somewhere else, or be more specific in naming or documentation.

src/window/window.rs
```diff
@@ -4,6 +4,10 @@ use sf_bool_ext::SfBoolExt;
 use system::{Vector2i, Vector2u};
 use window::{ContextSettings, Event, Style, VideoMode};
 
+/// The system native window handle type. Can be used to initialize SFMl within
```
crumblingstatue: `s/SFMl/SFML`

Wait... This doesn't sound right. I mean... We're not initializing SFML as a whole here, only a window, aren't we?

zstewar1: Yes. We initialize an SFML window based on an existing system window. I'm not sure of the internals, but I believe SFML auto-initializes other internals on an as-needed basis (e.g. connections to the X window server).

I'm not sure whether SFML takes control of the specified window handle or if it just initializes an OpenGL context so it can draw on it and otherwise leaves it alone. It might depend on whether you try to call other methods like set_title, set_size, etc. I haven't needed those in my use case (using SFML for an xsecurelock screensaver), so I don't know what happens if you call them (though I may experiment with them).

PR #196, opened 2018-12-15
Documentation: Add link to the wiki

fabienjuif:
Small change, I did this because this is not natural to look at wiki tab for me in github projects.
I am used to see documentation linked from the README.md (and linked to others MD files, but this is not the point here 😄 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,12 @@ Requirements
 
 Documentation
 =====
-The SFML official API (C++) is available here: https://docs.rs/sfml/0.14.0/sfml/
+The API documentation is available at: https://docs.rs/sfml/
 
-To reference and install `rust-sfml`, you can take a look at the [github wiki](https://github.com/jeremyletang/rust-sfml/wiki)\
+If you need help with setting up `rust-sfml` on your system, you can take a look at the [wiki](https://github.com/jeremyletang/rust-sfml/wiki).\
 Please take note that:
    * This wiki is supported by the community
-   * The `rust-sfml` core team doeesn't review it
+   * The `rust-sfml` core team doesn't review it
    * Your contribution is welcome
 
 License
--- a/README.md
+++ b/README.md
@@ -11,9 +11,15 @@ Requirements
 - [SFML 2.5](http://www.sfml-dev.org/download.php)
 - [CSFML 2.5](http://www.sfml-dev.org/download/csfml/)
 
-Usage
+Documentation
 =====
-You can find documentation about `rust-sfml` into the [dedicated wiki](https://github.com/jeremyletang/rust-sfml/wiki)
+The SFML official API (C++) is available here: https://docs.rs/sfml/0.14.0/sfml/
+
+To reference and install `rust-sfml`, you can take a look at the [github wiki](https://github.com/jeremyletang/rust-sfml/wiki)\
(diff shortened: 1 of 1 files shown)


fabienjuif:
Into the wiki:
 - I added a small part about MacOS
 - I move the Windows page to "Windows" (change title)
 - I added references about these page into the wiki home

README.md
```diff
@@ -11,6 +11,10 @@ Requirements
 - [SFML 2.5](http://www.sfml-dev.org/download.php)
 - [CSFML 2.5](http://www.sfml-dev.org/download/csfml/)
 
+Usage
```
crumblingstatue: 'Usage' makes it sound like the wiki is a comprehensive guide on how to use rust-sfml, but it's mostly a guide for setting up rust-sfml on different platforms if you're having difficulties doing so.

README.md
```diff
@@ -11,6 +11,10 @@ Requirements
...
+=====
+You can find documentation about `rust-sfml` into the [dedicated wiki](https://github.com/jeremyletang/rust-sfml/wiki)
```
crumblingstatue: Technically, the (API) "documentation" lives at https://docs.rs/sfml/0.14.0/sfml/ . Maybe we should also include that information, to avoid people thinking that the entire documentation is on the wiki.

Also, what do you mean by dedicated wiki? To me, the word dedicated here sounds like we made our own dedicated wiki instead of using GitHub's built-in wiki feature, which is not true.

README.md
```diff
@@ -11,9 +11,15 @@ Requirements
 - [SFML 2.5](http://www.sfml-dev.org/download.php)
 - [CSFML 2.5](http://www.sfml-dev.org/download/csfml/)
 
-Usage
+Documentation
 =====
-You can find documentation about `rust-sfml` into the [dedicated wiki](https://github.com/jeremyletang/rust-sfml/wiki)
+The SFML official API (C++) is available here: https://docs.rs/sfml/0.14.0/sfml/
```
crumblingstatue: This isn't the official C++ API. This is the actual Rust API of rust-sfml.
Also, we don't need to link to a specific version.

I would personally write this as

> The API documentation is available at: https://docs.rs/sfml/

README.md
```diff
@@ -11,9 +11,15 @@ Requirements
...
+
+To reference and install `rust-sfml`, you can take a look at the [github wiki](https://github.com/jeremyletang/rust-sfml/wiki)\
```
crumblingstatue: What do you mean here by 'reference'?

I would write this part as
> If you need help with setting up rust-sfml on your system, you can take a look at the [wiki](https://github.com/jeremyletang/rust-sfml/wiki).

README.md
```diff
@@ -11,9 +11,15 @@ Requirements
...
+Please take note that:
+   * This wiki is supported by the community
+   * The `rust-sfml` core team doeesn't review it
```
crumblingstatue: Typo (`s/doeesn't/doesn't/`)

README.md
fabienjuif: Yeah... I am very sorry

fabienjuif:
You are the one that have patience! 😅

Thank you for this review.

crumblingstatue:

PR #199, opened 2019-01-04
Create example with moving circle

DuckerMan:


crumblingstatue:
We usually add examples to showcase a specific feature of rust-sfml.
What feature of rust-sfml does this example showcase that other examples don't?

Also, you should make sure that your example compiles on latest version of the git master branch.
[As you can see on Travis](https://travis-ci.org/jeremyletang/rust-sfml/builds/475432716?utm_source=github_status&utm_medium=notification), it doesn't build.

crumblingstatue:
> OK, Which examples you need?

I don't have any in mind at the moment.

In the long term, it would be nice to port all the examples that the original C++ SFML has to Rust.

crumblingstatue:
Thank you for your effort, but I'm closing this, since I don't think we need such an example.
The pong example already covers what this example does.

I don't want to overload the examples directory with all kinds of overlapping examples. Quality over quantity.

But if you feel like you have an idea for an example that showcases something unique, feel free to suggest it!

PR #200, opened 2019-01-17
Fix error in window::Window example in doc

dmolina:
Example in https://docs.rs/sfml/0.14.0/sfml/window/struct.Window.html is not
working. 

It was using &Default::default(), and it is not working. However, 
&ContextSettings::default() is working. This change update that example in the documentation.

### DIFF
--- a/src/window/window.rs
+++ b/src/window/window.rs
@@ -26,12 +26,12 @@ pub type Handle = ffi::sfWindowHandle;
 /// # Usage example
 ///
 /// ```no_run
-/// use sfml::window::{Window, Event, Style};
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Can you clarify what you mean by "not working"? Does it result in a compile error? If so, what is the error message?

I tested the example on both rust-sfml 0.14 and the latest git master, and it compiles fine on both.

What Rust version are you using?

If possible, can you post the full code that demonstrates the error? (Pasting the example code into a main.rs doesn't compile, since it needs to be inside a function body)

PR #207, opened 2019-11-26
Implement VertexBuffer and its minimal example

silverweed:
I implemented the `VertexBuffer` struct from SFML 2.5.

All relevant methods and types are documented (the documentation is a close adaptation from the official SFML docs). 

Please let me know if something needs to change :)

### DIFF
--- a/src/graphics/vertex_buffer.rs
+++ b/src/graphics/vertex_buffer.rs
@@ -173,9 +173,13 @@ impl VertexBuffer {
     /// entities. It must be used only if you mix sf::VertexBuffer with OpenGL code.
     ///
     /// ```
-    /// let mut vb1 = VertexBuffer::new(/* ... */);
-    /// let mut vb2 = VertexBuffer::new(/* ... */);
-    /// ...
+    /// use sfml::graphics::{PrimitiveType, VertexBuffer, VertexBufferUsage};
+    ///
+    /// let mut vb1 = VertexBuffer::new(PrimitiveType::Triangles, 32, VertexBufferUsage::Static);
+    /// let mut vb2 = VertexBuffer::new(PrimitiveType::Quads, 12, VertexBufferUsage::Dynamic);
+    ///
+    /// // ...
+    ///
     /// VertexBuffer::bind(Some(&vb1));
     /// // draw OpenGL stuff that use vb1...
     /// VertexBuffer::bind(Some(&vb2));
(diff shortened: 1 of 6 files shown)


crumblingstatue:
Thank you for your contribution, I'll review it tomorrow!

crumblingstatue:
The doc-tests are failing, but CI didn't catch it, because it skips certain modules. I'll try and rework CI testing quickly. Meanwhile, try running cargo test locally and fixing it.

silverweed:
Sorry, I forgot to run `cargo test` before pushing. I fixed the failing doc test, now the should all pass.

src/graphics/vertex_buffer.rs
```diff
@@ -0,0 +1,246 @@
...
+    Stream = 0,
+    /// Occasionally changing data.
+    Dynamic = 1,
+    /// Rarely changing data.
+    Static = 2,
+}
+
+impl VertexBufferUsage {
+    pub(super) fn raw(self) -> sfVertexBufferUsage {
+        unsafe { ::std::mem::transmute(self) }
```
crumblingstatue: I'm sorry, this is nitpicking, but can't we do a simple `as` conversion here instead of unsafe `transmute`?
I know it shouldn't technically change anything, but unnecessary unsafe always bothers me. We should strive to avoid unsafe whenever possible.

silverweed: Nitpicking is a good habit :)
I removed the unnecessary unsafe code.

silverweed: Btw, isn't the same objection valid for all the other enums using `transmute` in their `raw` method? (e.g. `PrimitiveType`, `mouse::Button`, etc)

crumblingstatue:
Haha, the vertex buffer bind test is failing because of the headless CI, but I'll add a cfg to not run that test on CI.

silverweed:
Thank you for the review!

src/graphics/vertex_buffer.rs
crumblingstatue: > Btw, isn't the same objection valid for all the other enums using `transmute` in their `raw` method? (e.g. `PrimitiveType`, `mouse::Button`, etc)

That might be the case. If so, they should probably be changed to not use unsafe.

silverweed: Ok, I think I'll do another PR for that in the near future.

PR #210, opened 2020-01-02
make Rect::new, Vector2::new and Vector3::new const

silverweed:
The base constructors of `Rect` and `Vector2/3` can be made `const`, so I think it's nice to have them const. 
This allows defining compile-time constants of type `Vector` and `Rect`.

Probably this is also true for other types, but these are among the most commonly used, so we can start with just these.


crumblingstatue:
Looks good and makes sense, thank you!

PR #214, opened 2020-02-29
Add get/set_vertex methods to VertexArray and fix doc

silverweed:
The `VertexArray` struct currently has incorrect documentation on its `new` method: it states that its `vertex_count` argument is "The maximal number of vertex". However, that is really the *inital* number of vertices.

Moreover, there is no way to modify an existing Vertex: you can only `append` them - which resizes the array.

This PR addresses both problems by:
- fixing the documentation
- adding `get_vertex()`, `get_vertex_mut()` and `set_vertex()` methods to the `VertexArray`.

Edit:
- also adding `unchecked` versions of those methods.

### DIFF
--- a/src/graphics/vertex_array.rs
+++ b/src/graphics/vertex_array.rs
@@ -141,6 +141,13 @@ impl VertexArray {
         unsafe { &mut *(sfVertexArray_getVertex(self.vertex_array, index) as *mut _) }
     }
 
+    /// Sets the `index`-th vertex to `vertex`.
+    /// Panics if `index >= self.vertex_count()`.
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Looks good to me.
If you'd like, you could add some example usage to them in the vertex-arrays example.

silverweed:
@crumblingstatue I just realized I totally missed the Index and IndexMut traits of VertexArray. They do exactly the same as get/set_vertex, so maybe we should get rid of the checked version of get/set_vertex and keep only the _unchecked versions...

crumblingstatue:
Haha, I guess missed them too when reviewing your PR.
Yeah, that sounds good to me! We could change the Index impls to use get_unchecked internally.

PR #215, opened 2020-03-02
VertexArray: remove checked get/set_vertex, refactor Index* and example

silverweed:
Checked `get/set_vertex()` are superfluous since VertexArray already has Index/Mut traits. Leaving the unchecked versions, refactoring index/_mut to use them and updating the vertex array example.

### DIFF
--- a/examples/vertex-arrays.rs
+++ b/examples/vertex-arrays.rs
@@ -46,6 +46,35 @@ fn main() {
         vertex_array[1].color, vertex_array[1].position
     );
 
+    unsafe {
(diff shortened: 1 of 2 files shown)


silverweed:
Ping @crumblingstatue in case you missed it :)

crumblingstatue:
Ah, sorry, I totally forgot about it!

crumblingstatue:
I'm not at home, so I can't test it locally, but it looks good to me, so I'm merging it.

PR #224, opened 2020-05-23
add handle() method to Window and RenderWindow

silverweed:
This PR adds the ability to retrieve the SystemHandle from Window and RenderWindow.

### DIFF
--- a/src/graphics/render_window.rs
+++ b/src/graphics/render_window.rs
@@ -90,10 +90,10 @@ impl RenderWindow {
 
     /// Get the OS-specific handle of the window.
     ///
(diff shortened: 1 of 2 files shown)


crumblingstatue:
Looks good, but can you refer to `Handle` in the documentation instead of `sf::WindowHandle`?

crumblingstatue:
Also, maybe system_handle might be a better name for the method? It would be closer to the original `getSystemHandle` name.

silverweed:
Done!

crumblingstatue:
Thanks for your contribution!

Issue #226, opened 2020-09-01
Key/Button::is_pressed are not thread safe


crumblingstatue:
At least on X11, it can trigger all kinds of errors to call them from a different thread.
And someone messaged me that it can cause segfaults for them to do so.

memoryleak47:
One partial solution to this would to add a dummy &RenderWindow parameter to Key::is_pressed. This should prevent it from being called in another thread than where the RenderWindow lives, because RenderWindow is !Sync.

This wouldn't work whenever the user creates multiple RenderWindows though.

crumblingstatue:
Looks like is_pressed isn't the only misbehaving API. There can be threading related errors for example if you create 2 sf::RenderWindows on 2 different threads. And there is probably more.

I don't have any clear plan on how to solve all of these threading issues, so for the time being I just [added a warning to the documentation](https://github.com/jeremyletang/rust-sfml/commit/9e00ede4d9585f99a796bfcbe6d2a9e6109be642).

memoryleak47:
What about adding a `static unique_threadid: OnceCell<ThreadId>` containing the thread id of the thread in which a RenderWindow is spawned.
And in `RenderWindow::new()` one
- sets `unique_threadid` to `thread::current().id()` if `unique_threadid.get().is_none()`
- does nothing if `unique_threadid.get() == Some(thread::current().id())`
- returns an error / panics, if the `unique_threadid.get().is_some()` but contains another thread-id than `thread::current().id()`.

Using this, every RenderWindow has to live in the same thread.

This then allows to either:
- let non-thread-safe functions panic / error if they are called from the wrong thread (by comparing `thread::current().id()` to `unique_threadid`)
- add a dummy `&RenderWindow` argument to every non-thread safe function, which implies that this function is only called from the `unique_threadid`-thread.

crumblingstatue:
That sounds like it might work. If you'd like you can submit a pull request, but for now I'll try to explore what other thread safety issues we might have.

PR #227, opened 2020-09-07
[WIP] thread safety

memoryleak47:
This adds the `thread_safety::assert_window_thread`-function, which panics if called from some other thread than the window-thread.
There is still a lot of room for experimentation in order to find out what functions should call `thread_safety::assert_window_thread`, or whether this mechanism is too restrictive or relaxed.
cc #226

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,9 @@ audio = ["csfml-audio-sys"]
 # Used to skip running certain tests on CI, since it's running in a headless environment.
 ci-headless = []
 
+[dependencies.once_cell]
+version = "1.4"
+
 [dependencies.widestring]
 version = "0.4.0"
 optional = true
--- a/src/graphics/render_window.rs
+++ b/src/graphics/render_window.rs
@@ -6,6 +6,7 @@ use crate::{
     },
     sf_bool_ext::SfBoolExt,
     system::{SfStrConv, Vector2f, Vector2i, Vector2u},
+    thread_safety,
     window::{ContextSettings, Cursor, Event, Handle, Style, VideoMode},
 };
 use csfml_system_sys::*;
@@ -47,6 +48,8 @@ impl RenderWindow {
         style: Style,
         settings: &ContextSettings,
     ) -> RenderWindow {
+        thread_safety::set_window_thread();
+
         title.with_as_sfstr(|sfstr| {
             let sf_render_win: *mut ffi::sfRenderWindow = unsafe {
                 ffi::sfRenderWindow_createUnicode(
@@ -80,6 +83,8 @@ impl RenderWindow {
     /// * settings - Additional settings for the underlying OpenGL context
     #[must_use]
     pub unsafe fn from_handle(handle: Handle, settings: &ContextSettings) -> RenderWindow {
+        thread_safety::set_window_thread();
+
         let sf_render_win: *mut ffi::sfRenderWindow =
             ffi::sfRenderWindow_createFromHandle(handle, &settings.raw());
         assert!(!sf_render_win.is_null(), "Failed to create Window");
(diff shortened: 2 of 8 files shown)


crumblingstatue:
This simple example still produces varying errors
```Rust
    std::thread::spawn(|| {
        let mut rw = RenderWindow::new((640, 480), "Hi", Style::CLOSE, &Default::default());
    });
    println!("{}", Key::K.is_pressed());
```

This is likely because is_pressed() also inits some kind of windowy context (opens a connection to the X server on Linux at least), so it really should also *set* the window thread as well. Probably almost everything should set the window thread instead of just merely asserting it.

crumblingstatue:
Do you have any objection to just getting rid of assert_window_thread entirely and just using set_window_thread everywhere?
```diff
diff --git a/src/thread_safety.rs b/src/thread_safety.rs
index 96c1e66..894cb77 100644
--- a/src/thread_safety.rs
+++ b/src/thread_safety.rs
@@ -12,20 +12,8 @@ pub(super) fn set_window_thread() {
         None => WINDOW_THREAD.set(current_id).unwrap(),
         Some(id) => {
             if *id != current_id {
-                panic!("Opening Windows in multiple threads is not supported.")
+                panic!("Graphics/windowing functionality is only supported on the same thread.")
             }
         }
     }
 }
-
-// asserts that the either current thread is WINDOW_THREAD, or that WINDOW_THREAD is unset
-pub(super) fn assert_window_thread(func_name: &str) {
-    if let Some(&id) = WINDOW_THREAD.get() {
-        if id != current().id() {
-            panic!(
-                "{} has to be called in the same thread as the RenderWindow!",
-                func_name
-            );
-        }
-    }
-}
diff --git a/src/window/keyboard.rs b/src/window/keyboard.rs
index 0da0417..b5c3552 100644
--- a/src/window/keyboard.rs
+++ b/src/window/keyboard.rs
@@ -127,7 +127,7 @@ impl Key {
     /// triggered.
     #[must_use]
     pub fn is_pressed(self) -> bool {
-        thread_safety::assert_window_thread("Key::is_pressed()");
+        thread_safety::set_window_thread();
 
         unsafe { ffi::sfKeyboard_isKeyPressed(self as _) }.to_bool()
     }
diff --git a/src/window/mouse.rs b/src/window/mouse.rs
index ca9aa57..b4b440c 100644
--- a/src/window/mouse.rs
+++ b/src/window/mouse.rs
@@ -77,7 +77,7 @@ impl Button {
     /// triggered.
     #[must_use]
     pub fn is_pressed(self) -> bool {
-        thread_safety::assert_window_thread("Button::is_pressed()");
+        thread_safety::set_window_thread();
 
         unsafe { ffi::sfMouse_isButtonPressed(self.raw()) }.to_bool()
     }
diff --git a/src/window/video_mode.rs b/src/window/video_mode.rs
index 6286ea4..9495f8f 100644
--- a/src/window/video_mode.rs
+++ b/src/window/video_mode.rs
@@ -79,7 +79,7 @@ impl VideoMode {
     /// return the urrent desktop video mode
     #[must_use]
     pub fn desktop_mode() -> Self {
-        thread_safety::assert_window_thread("VideoMode::desktop_mode()");
+        thread_safety::set_window_thread();
 
         unsafe { Self::from_raw(ffi::sfVideoMode_getDesktopMode()) }
     }
```

crumblingstatue:
I meant you could do it, unless you want me to make another pull request based on this, and close this one.

crumblingstatue:
Thank you, I'll merge this as a good first step towards thread safety

memoryleak47:
Merging this is maybe a little early;
As we are only certain that this problem arises when used with X11, I could image that a fair amount of non-X11 users which use multi-threading will now have their code broken.
Maybe we should add `#[cfg(target_os = "linux")]`.

crumblingstatue:
That would mean that the application they made would work on non-linux platforms, but panic on Linux. Shouldn't the panic be universal so they design their application in a way that works on all platforms?

PR #229, opened 2020-10-04
Add derive `Hash` to joystick axis

datMaffin:
The joystick axis struct is similar to the keyboard key struct which
already derives `Hash`


crumblingstatue:
Makes sense, thank you for your contribution!

PR #231, opened 2020-11-25
Implement conversion from Vector<T> to Vector<U>

### LINKED ISSUE #230: From<Vector<_>> for Vector<_> structs
Implement From<Vector<_>> for each vector type so that it is possible to do i.e. `Vector2f::from(vec)`. I keep writing `Vector2f { x: vec.x as f32, y: vec.y as f32 }` to convert from integer to float vectors and it's a pain.

alexdevteam:
Adds the From<Vector<T>> trait to Vector2<U> and Vector3<U>. With it, it is now possible to do stuff like `Vector2i::from(vec) // vec is a Vector2f`. Closes #230.

### DIFF
--- a/src/system/vector2.rs
+++ b/src/system/vector2.rs
@@ -99,6 +99,26 @@ impl_ops!(Sub, sub, i32, u32, f32);
 impl_ops!(Mul, mul, i32, u32, f32);
 impl_ops!(Div, div, i32, u32, f32);
 
+macro_rules! impl_conv {
+    ( $($_type:ty => $($_conv_type:ty),+);+ $(;)?) => {
+        $(
+            $(
+                impl From<Vector2<$_conv_type>> for Vector2<$_type> {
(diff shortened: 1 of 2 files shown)


crumblingstatue:
Hi!
However, From should not be implemented for conversions that can fail. See the documentation on the From trait.
f32 to i32 can fail for example, and others can too in your code.
However, I would accept an implementation of From<T> for Vector2<U> where U: From<T>.
I might also accept TryFrom<T> for Vector2<U> where U: TryFrom<T>, if that's something that you would want.
If someone needs a more convenient form for lossy conversion, it shouldn't be too hard to write an extension trait for that, but I don't want to encourage it in the main library.

crumblingstatue:
Since there has been no activity for 2 weeks and I don't want to merge this as-is, I'm closing it, but feel free to open another pull request if you would like to implement TryFrom or some other means of lossy conversion.

PR #233, opened 2021-02-04
Made a small part of the code a wee bit clearer

actuday6418:


examples/pong.rs
```diff
@@ -91,7 +91,7 @@ fn main() {
     let mut is_playing = false;
 
     loop {
-        while let Some(event) = window.poll_event() {
+        for event in window.poll_event() {
```
crumblingstatue, line 94: This doesn't loop over all events.
It only handles a single event.

Option implements IntoIterator, which is why this code is possible.
It will yield a single item if it's Some, and nothing if it's None.

But what we want is poll events *until* there is none left, not a single event if  there is one.

crumblingstatue (commented):
Hi!

Thank you for being interested in contributing!

However, I have to reject this pull request. See my comment on the code.

examples/pong.rs
actuday6418: Thank you! I was wondering why my program appeared to be handling just one event each frame. 

I noticed also that some of the examples have code that's not compatible with the latest version of rust-sfml.
For example, `Key::SPACE` in examples/pong.rs needs to be replaced with `Key::Space` . Could I make the obvious changes so the examples compile and run with the latest release?

crumblingstatue: You are looking at the example code on the master branch, which is actually the latest version. If you wanna look at the examples of the latest release, you could select the tag of it on GitHub and view that, or "view source code" on docs.rs, and look at the examples folder.

Issue #239, opened 2021-07-15
Music::from_stream and from_memory are unsound


Kolsky:
Got the following message in debug mode after creating the music with helper function, which has a shorter scope for its Vec/Cursor, and trying to play it afterwards.
error: process didn't exit successfully: `target\debug\sfml_test.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION).
Supposedly, it's CSFML/SFML fault for no indication of pointer's lifetime in documentation or elsewhere. Though I don't have a clear solution. At least `impl Into<Box<[u8]>>` with pinning should suffice for from_memory.

crumblingstatue:
Can you please provide a self-contained minimal example against the master branch that exhibits the crash, along with a stack trace?

crumblingstatue:
Alright, I understand the problem now.
I think I'll solve this by making a public InputStream type instead of the current T: Read + Seek requirement.

crumblingstatue:
Hi! Can you test out this branch? https://github.com/jeremyletang/rust-sfml/tree/public-inputsream
Let me know if it fixes the issue!

Kolsky:
Sorry for not replying, I'll try it in the meantime

Kolsky:
It doesn't seem to resolve the issue. Also the CSFML creates its own `CallbackStream` when calling `sfMusic* sfMusic_createFromStream(sfInputStream* stream)` with `CallbackStream(sfInputStream* stream) : myStream(*stream) { }`, if I understand the code correctly. Then the requirement is still the same as in pull request, I guess? Also parametrizing over just a lifetime over the type might be preferable for convenience.
Maybe we should ask CSFML/SFML owners about this.

crumblingstatue:
Can you please supply a minimal test case that reproduces the problem on master? I really don't understand the problem.

Kolsky:
There isn't really much else to add.
A minimal example with public-inputstream branch: https://github.com/Kolsky/sfml_open_music
Call stack of the Main Thread:
```
ntdll.dll!00007ffdf43aae14() (Unknown Source:0)
KernelBase.dll!00007ffdf1c626ee() (Unknown Source:0)
csfml-audio-2.dll!00007ffdc996368b() (Unknown Source:0)
csfml-audio-2.dll!00007ffdc9957cde() (Unknown Source:0)
csfml-audio-2.dll!00007ffdc9958318() (Unknown Source:0)
[...]

PR #240, opened 2021-07-15
Fixed unsoundness in Music constructors

Kolsky:
Has a tiny amount of breaking changes, has not tested yet
#239


Kolsky:
Update: memory impl works fine, but the stream one fails with either STATUS_STACK_BUFFER_OVERRUN or the other one.

Kolsky:
Still doesn't quite gets it. At least STATUS_ACCESS_VIOLATION seems to be gone, maybe there is something else with the library or my implementation.

crumblingstatue:
Thank you for your work, I'll review it tomorrow

Kolsky:
Here is my stack trace after calling `ffi::sfMusic_destroy` with a stream, just in case
image

src/audio/music.rs
```diff
@@ -119,13 +127,17 @@ impl Music {
     ///
     /// [`play`]: Music::play
     #[must_use]
-    pub fn from_memory(mem: &[u8]) -> Option<Music> {
-        let music_tmp =
-            unsafe { ffi::sfMusic_createFromMemory(mem.as_ptr() as *const _, mem.len()) };
+    pub fn from_memory(mem: impl Into<Box<[u8]>>) -> Option<Music> {
```
crumblingstatue, line 130: To be honest, I would much rather stick to some kind of borrowing if possible, even if it involves lifetimes.
The original SFML API borrows here too, there is no keep-alive magic.

Kolsky: It's probably feasible, however the InputStream is used in many places within the api, so designing the solution would take a bit longer, and I probably won't be able to test everything. Should I try it?

Kolsky: That's actually kind of interesting. The Font and the Music are relying on having their data to be kept alive, but only the Music does have methods which start with open, Font, on the other hand, has a warning instead.

crumblingstatue: Does InputStream need to be changed at all?
Music::from_stream takes &T, which can have a lifetime associated with it.
I made an experimental branch here:
https://github.com/jeremyletang/rust-sfml/tree/music-memsrc
Let me know what you think, and if it solves your case.

Kolsky: Thanks, I'll see what I can do from here.

Kolsky: There is still the need to pin an InputStream in a memory and keep it alive somehow for the lifetime of a Music. Box it and put in the field, perhaps?

crumblingstatue: Is it required for soundness, or is it required for your use case because you want to store it along with the music?
If it's required for soundness, can you explain further? Because I think borrowing it should be enough for soundness.

If it's the latter, there could be a special Music type that stores the stream alongside the sfMusic handle.

Kolsky: I think it's required, but not enough for soundness, as can be seen from documentation.
image
There is certainly more to it due to STATUS_STACK_BUFFER_OVERRUN after calling sfMusic_destroy with the stream opened, so I don't want to insist on adding a box just for the sake of it, but it'd likely to become the necessary part of a fix.

crumblingstatue:
Can you post code that reproduces the error with my branch? I don't understand what's causing the problem, so it would be nice to have an example.

I'm going to bed now, but I'll continue investigating this tomorrow.
Thank you for your hard work!

Kolsky:
https://github.com/Kolsky/sfml_open_music
Oddly enough, there is no need for the box, maybe C interface does some shenanigans? Idk what's real anymore :(
Btw, it does reproduce the bug, so feel free to investigate.

crumblingstatue:
Hmmm.... Unfortunately, I can't seem to reproduce the bug with that example. I tried all the `Load` variants, and run the program under valgrind, but it didn't find anything. But I'll try investigating further.

crumblingstatue:
I fixed a bug in inputstream. Let me know if the issue still happens.
https://github.com/jeremyletang/rust-sfml/commit/44dd1eb4630fe4af24d965ab6166da8de4c4b966

Kolsky:
To be honest, it was already patched out in the experimental branch. Could've solved the case with ACCESS_VIOLATION, but unfortunately it's not the source of this particular behaviour.
image
I'm under stable-x86_64-pc-windows-msvc toolchain with SFML and CSFML 2.5 x64.

src/audio/music.rs
crumblingstatue: Alright, I understand why we need to keep the stream in memory now.

Sorry for taking so long to understand it. I'll think about how to best solve this, but I'll consider your OwnPtr solution too.

crumblingstatue:
https://github.com/jeremyletang/rust-sfml/commit/e4f1149554735f87471c75cbb2424d81f05de678 should have fixed the STATUS_STACK_BUFFER_OVERRUN problem.
But let me know if you have additional problems, the stream API might not be safe just yet.

crumblingstatue:
A lot of things have changed since this pull request was made, so I'm closing it.
But thank you very much for your work!
Don't take this as discouragement to submit future PRs!

PR #244, opened 2021-09-11
rewritten cursors.rs for the new version

linux-admin0001:
Fixed these errors:
image


crumblingstatue:
Hi!
What you are looking at is examples from the latest git commit, but you are trying to build those from a stable version.
The examples are correct for the latest git version. You should instead take a look at the [examples for the stable version](https://github.com/jeremyletang/rust-sfml/blob/v0.16.0/examples/cursor.rs)

Issue #251, opened 2021-11-26
Convex Shape points not accurate 


DanielLi156:
I've been using the Convex Shapes to create a triangle

```rust
let mut shape2 = ConvexShape::new(3);
shape2.set_point(0, Vector2f::new(0.0, 0.0));
shape2.set_point(1, Vector2f::new(0.0, 100.0));
shape2.set_point(2, Vector2f::new(100.0, 0.0));

for point in shape2.points() {
    println!("x:{} y:{}", point.x, point.y);
}
```

However when looping through the points and printing them out sometimes the points become weird numbers. I have no idea how to fix it either.

```rust
x:0 y:100
x:100 y:0
x:434583.44 y:-0.000000000000000000000000000000000000000047353
```

PR #252, opened 2021-11-27
Post-increment index in ConvexShapePoints iterator

kmdreko:
The iterator for `ConvexShape::points()` would increment the current
index *first* and use that to access the point, which causes it to skip
the first one and result in an invalid read for the last one. This
changes it to increment the index *after* getting the point at that
index.

Fixes https://github.com/jeremyletang/rust-sfml/issues/251

### DIFF
--- a/src/graphics/convex_shape.rs
+++ b/src/graphics/convex_shape.rs
@@ -260,13 +260,14 @@ impl Iterator for ConvexShapePoints {
         if self.pos == point_count {
             None
         } else {
-            self.pos += 1;
-            unsafe {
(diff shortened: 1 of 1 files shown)


kmdreko:
Changes modeled after implementation in [`Iterator for sfVideoModeVectorIter`](https://github.com/jeremyletang/rust-sfml/blob/19b5e1082fbb58d7c1646eef32cb6fd01353c9d3/src/ffi/window.rs#L444-L456)

crumblingstatue:
Thank you!

PR #260, opened 2022-02-05
Update Event

linux-admin0001:
Update Event for use in HashMap


crumblingstatue:
This unfortunately isn't feasible because of the floating point values inside certain events, but for most purposes, std::mem::discriminant can be used to put event discriminants in a HashMap.

PR #263, opened 2022-05-28
Added simple type conversions for rect, vector2, and vector3.

dogunbound:
I feel that a simple type conversion is needed between:
IntRect and FloatRect
Vector2i, Vector2f, and Vector2u
Vector3i, and Vector3f

Also, simply warn the user in documentation that conversions that don't "fit" (i.e. f32::MAX doesn't convert nicely into i32) will silently fail with no warning. Let the user worry about what datatype fit into where.

### DIFF
--- a/src/graphics/rect.rs
+++ b/src/graphics/rect.rs
@@ -125,6 +125,17 @@ impl IntRect {
             height: raw.height,
         }
     }
+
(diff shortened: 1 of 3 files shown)


dogunbound:
Gonna rewrite it with macros instead. Didn't know I about them.

PR #264, opened 2022-05-29
Added primitive number data type conversion for vec2, vec3, and Rect.

dogunbound:
Easy conversions between all numeric primitive data types for vec2, vec3, and
Rect are now implemented with either lossy_from/lossy_into,
try_from/try_into, or from/into.

lossy_from/lossy_into is performed when converting between floating
point and integers. This case silently fails on overflow / etc.. Hence
"lossy".

try_from/try_into returns a result on with passing being the vector with
the preferred data type, and failing returning a TryFromIntError.

from/into are performed on cases where the into wraps the from entirely
and is completely safe to convert between the two. These special cases
are conversions from a unsigned integer into a signed integer with more
bits. (i.e. u8 is safe to convert into i16, i32, i64, and i128).

### DIFF
--- a/src/graphics/rect.rs
+++ b/src/graphics/rect.rs
@@ -58,9 +58,9 @@ macro_rules! lossy_from_impls {
         }
     )*}
 }
-
+// float -> float (differently size)
 lossy_from_impls!(f32, f64; f64, f32);
-
+// f32 -> integer
 lossy_from_impls!(
 f32, i8;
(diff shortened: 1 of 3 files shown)


dogunbound:
cargo test passes.

dogunbound:
Forgot signed to signed and unsigned to unsigned integer conversions.

PR #265, opened 2022-05-29
Added primitive number data type conversion for vec2, vec3, and Rect.

dogunbound:
Easy conversions between all numeric primitive data types for vec2, vec3, and
Rect are now implemented with either lossy_from/lossy_into,
try_from/try_into, or from/into.

lossy_from/lossy_into is performed when converting between floating
point and integers. This case silently fails on overflow / etc.. Hence
"lossy".

try_from/try_into returns a result on with passing being the vector with
the preferred data type, and failing returning a TryFromIntError.

from/into are performed on cases where the into wraps the from entirely
and is completely safe to convert between the two. These special cases
are conversions from a unsigned integer into a signed integer with more
bits. (i.e. u8 is safe to convert into i16, i32, i64, and i128).

LossyFrom and LossyInto are copy and pasted because I don't want the user to have to import a "lossy conversion" crate to be able to use lossy conversions. Simply importing vec2, vec3, or rect will automagically give them lossy conversions this way.

The reason why try_from requests for an error data type rather than defaulting to TryFromIntError in the macro definition is because try_from doesn't indicate that it is for integers. I feel it's best if I'm not explicit of the type in the function declaration, then don't put specific things in the macro.

### DIFF
--- a/src/graphics/rect.rs
+++ b/src/graphics/rect.rs
@@ -58,9 +58,9 @@ macro_rules! lossy_from_impls {
         }
     )*}
 }
-
+// float -> float (differently size)
 lossy_from_impls!(f32, f64; f64, f32);
-
+// f32 -> integer
 lossy_from_impls!(
 f32, i8;
 f32, i16;
@@ -73,7 +73,10 @@ f32, u16;
 f32, u32;
 f32, u64;
 f32, u128;
-f32, usize;
+f32, usize
+);
+// integer -> f32
+lossy_from_impls!(
 i8, f32;
 i16, f32;
 i32, f32;
@@ -87,7 +90,7 @@ u64, f32;
 u128, f32;
 usize, f32
 );
-
+// f64 -> integer
 lossy_from_impls!(
 f64, i8;
 f64, i16;
@@ -100,7 +103,10 @@ f64, u16;
 f64, u32;
 f64, u64;
 f64, u128;
-f64, usize;
+f64, usize
+);
+// integer -> f64
+lossy_from_impls!(
 i8, f64;
 i16, f64;
 i32, f64;
@@ -125,7 +131,7 @@ macro_rules! from_impls {
         }
     )*}
 }
-
+// unsigned -> signed
 from_impls!(
 u8, i16;
 u8, i32;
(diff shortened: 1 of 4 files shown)


dogunbound:
Something I want you to consider if you accept these changes, we could implement lossy conversions for ALL primitive datatypes.

If you think that's a decent idea, then lmk.

dogunbound:
One more thing, If you do accept these changes, is it possible for you to update rust-sfml's crate?

crumblingstatue:
Thank you for your submission! I'll review it soon! (might only have time tomorrow)
> One more thing, If you do accept these changes, is it possible for you to update rust-sfml's crate?
rust-sfml is kind of in a limbo right now. I started doing a big CSFML rewrite months ago, but I lost motivation.
Right now, among other things, rust-sfml requires clang to build, which could be a pain for users.
I'll try getting it into a usable state, but I can't promise anything because I'm working on a lot of projects in parallel right now.

dogunbound:
How could I reference the github rather than whats stable through the crate? Is that even possible?

dogunbound:
> Thank you for your submission! I'll review it soon! (might only have time tomorrow)
> 
> > One more thing, If you do accept these changes, is it possible for you to update rust-sfml's crate?
> 
> rust-sfml is kind of in a limbo right now. I started doing a big CSFML rewrite months ago, but I lost motivation. Right now, among other things, rust-sfml requires clang to build, which could be a pain for users. I'll try getting it into a usable state, but I can't promise anything because I'm working on a lot of projects in parallel right now.

How could I reference the github rather than whats stable through the crate? Is that even possible?

crumblingstatue:
> How could I reference the github rather than whats stable through the crate? Is that even possible?

That's very simple

```toml
[dependencies.sfml]
git = "https://github.com/jeremyletang/rust-sfml.git"
```

src/graphics/rect.rs
```diff
@@ -22,6 +26,267 @@ pub type IntRect = Rect<i32>;
 /// A [`Rect`] of `f32`.
 pub type FloatRect = Rect<f32>;
 
+pub trait LossyFrom<T> {
+    fn lossy_from(v: T) -> Self;
```
crumblingstatue, line 30: I don't think we need traits for this. We could just have inherent methods called `as_f32`, `as_f64`, `as_i8`, etc.
That way we don't pollute the API with traits.

Also it looks like these traits are pasted into each module (rect, vector2, vector3), which is definitely not ideal.

PR #266, opened 2022-05-31
Removed PartialOrd and Ord from vectors

dogunbound:
These are not implemented in SFML 3, therefore should not be implemented here.


dogunbound:
WTF? 4 files changed?

PR #269, opened 2022-06-01
fetching position and size implemented for Rect

dogunbound:
This feature is not implemented for Rust 2.5, but it is on their main
repository. I thought it was a convenient feature to have regardless of
not being on SFML stable.

Pretty sure it will be out for sfml3


crumblingstatue:

PR #270, opened 2022-06-03
Vertex buffer new and clone now return with errors instead of panicing.

dogunbound:
The user should be the ones that define when and where the API crashes.
A panic in the API is unwarrented.

The added package (anyhow) is for easy error handling / messaging. This package
is used by many people.

### DIFF
--- a/src/graphics/vertex_buffer.rs
+++ b/src/graphics/vertex_buffer.rs
@@ -1,5 +1,5 @@
-use anyhow::{Context, Result};
 use std::ptr::NonNull;
+use thiserror::Error;
 
 use crate::{
     ffi::graphics::*,
@@ -31,6 +31,12 @@ pub struct VertexBuffer {
     vertex_buffer: NonNull<sfVertexBuffer>,
 }
 
+#[derive(Debug, Error)]
+pub enum VertexBufferError {
+    #[error("`{0}`")]
+    Message(String),
+}
+
 impl VertexBuffer {
     /// Create a new initialized vertex buffer
     ///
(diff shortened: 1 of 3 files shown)


src/graphics/vertex_buffer.rs
```diff
@@ -224,16 +254,6 @@ impl VertexBuffer {
     }
 }
 
-impl Clone for VertexBuffer {
```
crumblingstatue, line 227: I don't see the need to remove the original panicking Clone implementation for `VertexBuffer`.
It's there as an option if you don't care about handling a clone failure.

src/graphics/vertex_buffer.rs
```diff
@@ -30,6 +31,12 @@ pub struct VertexBuffer {
     vertex_buffer: NonNull<sfVertexBuffer>,
 }
 
+#[derive(Debug, Error)]
+pub enum VertexBufferError {
+    #[error("`{0}`")]
+    Message(String),
```
crumblingstatue, line 37: Holding a String is a bit heavy here, since this is just two statically determinable errors.
I would personally create two types here.
```Rust
#[derive(Debug, Error)]
#[error("failed to create vertex buffer")]
pub struct VertexBufferCreateError;

#[derive(Debug, Error)]
#[error("failed to clone vertex buffer")]
pub struct VertexBufferCloneError;
```

crumblingstatue:
Since C++ `VertexBuffer` copy constructor has an infallible API, I don't see much need to try and handle any errors here, since there is not much we can do, and it's extremely unlikely that an error would happen here.

PR #271, opened 2022-06-04
Update documentation to warn users not to store sprites and text in a struct

dogunbound:


src/graphics/sprite.rs
```diff
@@ -12,6 +12,11 @@ use std::{
     ptr::{self, NonNull},
 };
 
+/// !! WARNING !!
```
crumblingstatue, line 15: I wouldn't put such a dramatic warning here.

Maybe `Note:` would be better.

src/graphics/sprite.rs
```diff
@@ -12,6 +12,11 @@ use std::{
...
+/// Currently there is no way to store sprites in a struct. All sprites must be made right before
```
crumblingstatue, line 16: It's not impossible to store them in a struct, it's just not very practical.
I would rephrase this as
> Currently, it is not feasible to store sprites long term.
> A common pattern with rust-sfml is to create a `Sprite` right before you start drawing, and draw all the sprites you want with it. You can change its properties using `set_texture`, `set_position`, etc., before drawing it, as many times as you need to.

src/graphics/sprite.rs
dogunbound: Changed to Note:

src/graphics/sprite.rs
dogunbound: I copy and pasted that for sprite.rs

text.rs you may want to read.

PR #274, opened 2022-06-25
Added deserializer to rect

dogunbound:
Saw deserializer added for Vector. Thought it should go with Rect. My reasoning is that I want to save Rects in JSONs for UI and other niceties.

Also, it is good to be consistent. If Vector2 and Vector3 have this functionality, so should Rect, since it is also a simple data type.


crumblingstatue:

PR #277, opened 2022-07-24
Implement vector operations that are in main SFML Vector2.ini

dogunbound:
Here is a link to file we need to match:
https://github.com/SFML/SFML/blob/master/include/SFML/System/Vector2.inl

Mul, and Div impl needed to be changed to match Main SFML a bit more.
Added cwise_mul and cwise_div because that was what Div and Mul
originally inteded to do.

Added dot, perpendicular, length_sq, and cross because they were simple
to implement, and I believe they should be in rust-sfml also.

### DIFF
--- a/examples/pong.rs
+++ b/examples/pong.rs
@@ -54,7 +54,7 @@ fn main() {
 
     // Create the left paddle
     let mut left_paddle = RectangleShape::new();
-    left_paddle.set_size(paddle_size - 3.);
+    left_paddle.set_size(paddle_size - Vector2f::new(3., 3.));
     left_paddle.set_outline_thickness(3.);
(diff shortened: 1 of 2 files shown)


dogunbound:
If you accept this, I'll also work on Vector3. If you want to put it in the same PR, lmk and I'll do that. Otherwise, just accept it and you'll see Vector3 in another PR.

crumblingstatue:
Thanks, it looks good to me, although I'm not good at math.

I don't personally use Vector types for math very much, but I don't see the harm in adding a few mathematical methods.

dogunbound:
@crumblingstatue 

I didn't add it because of just math. I added it because main SFML had these changes.

PR #279, opened 2022-07-25
Added serializer for SFML keys, buttons, and mouse wheel

dogunbound:
I was trying to add custom control bindings for my app. I noticed that
if the Key, Button, and MouseWheel enums derived Serializer and Deserializer, it would help ALOT
  
  
  I understand if you don't accept this PR. This is quite niche, and I could do a custom serializer for this.


crumblingstatue:
I'm merging this, with the caveat that I reserve the right to change these types in a way that breaks backwards compatibility with previously serialized data. No backwards compatibility guarantees for serialized rust-sfml types.

PR #286, opened 2022-08-09
Make enum CursorType Hashable

dogunbound:
My inspiration for doing this is to allow me to create a hashmap of cursors
in my resource manager, with relevant keys rather than using custom
keys.


crumblingstatue:
Makes sense to impl Hash for this, thanks!

PR #288, opened 2022-08-15
Added more doc examples for the render_window module

dogunbound:
The most important documentation for it would be how to handle setting
cursors


crumblingstatue:
Thanks, I'll check it out tomorrow

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Examples for Vector3
 - Examples for Vector2
 - Examples for Rect
+- Examples for Render Window
```
crumblingstatue, line 9: Nitpick: Should be RenderWindow

src/graphics/render_window.rs
```diff
@@ -167,6 +268,24 @@ impl RenderWindow {
...
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// while window.is_open() {
+    ///     // Do something
+    ///     window.close();
```
crumblingstatue, line 284: I think this is kind of a silly example, because you close it right away on the first iteration.
I think you should close on `Closed` event, even if it makes the example longer.

src/graphics/render_window.rs
```diff
@@ -167,6 +268,24 @@ impl RenderWindow {
...
+    /// }
+    ///
+    /// // Once window is closed, we can do other thing.
```
crumblingstatue, line 287: Spelling: s/thing/things/

src/graphics/render_window.rs
```diff
@@ -190,6 +324,21 @@ impl RenderWindow {
...
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// while window.is_open() {
+    ///     // Draw something
+    ///     window.display();
```
crumblingstatue, line 339: You should at least clear with a color here, or else it will just display garbage.

src/graphics/render_window.rs
```diff
@@ -258,6 +474,19 @@ impl RenderWindow {
...
+    ///
+    /// ```no_run
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.set_mouse_cursor_visible(true);
```
crumblingstatue, line 488: You're calling `set_mouse_cursor_visible` here, not `set_mouse_cursor_grabbed`

src/graphics/render_window.rs
```diff
@@ -289,6 +530,18 @@ impl RenderWindow {
...
+    ///
+    /// ```no_run
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.set_key_repeat_enabled(true);
```
crumblingstatue, line 543: I think my problem with a lot of these examples is that they are so simple, they are completely self-explanatory, and redundant.
It's already very apparent from the function signature that you can call it with either `false` or `true`, it really needs no dedicated example.
Adding tons of small examples here with a lot of setup and teardown (creating `RenderWindow`, etc.), just makes testing take a lot longer than it needs to be. Costing more time and energy.

I honestly believe small self-explanatory examples like this are unnecessary and can be removed.

src/graphics/render_window.rs
```diff
@@ -308,6 +561,18 @@ impl RenderWindow {
...
+    ///
+    /// ```no_run
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.set_active(true);
```
crumblingstatue, line 574: Same here as point above, etc.

src/graphics/render_window.rs
```diff
@@ -328,6 +603,19 @@ impl RenderWindow {
...
+    /// ```no_run
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # use sfml::system::Vector2;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// assert_eq!(window.position(), Vector2::new(0, 0));
```
crumblingstatue, line 617: This is wrong. The window's position is not guaranteed to be `(0, 0)` when it's created.
The only reason this doesn't fail is because it's `no_run`.
Assertions in a `no_run` test are useless. In fact, worse than useless because they create misinformation, like in this case.

You should remove all assertions from `no_run` tests.

src/graphics/render_window.rs
```diff
@@ -408,6 +778,19 @@ impl RenderWindow {
...
+    ///
+    /// ```no_run
+    /// # use sfml::window::{Event, Style};
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// assert_eq!(window.has_focus(), false);
```
crumblingstatue, line 792: This is wrong, like the example above. A window is not guaranteed to not have focus when it's created.
Remove this assertion.

crumblingstatue (changes requested):
Sorry for being harsher than usual, I just don't want to pollute the documentation with examples that are not useful.

src/graphics/render_window.rs
dogunbound: I was debating this in my head. I decided to look at the official rust docs and they used small redundant examples. I just copied what the official rust docs did.

I kind of agree with you though. I will get rid of the small self-explanatory examples then.

src/graphics/render_window.rs
dogunbound: What about set_position's example? We CANNOT run it because CI doesn't have x11, but it does work. What do you want to do about this example?

src/graphics/render_window.rs
dogunbound: done

src/graphics/render_window.rs
dogunbound: This is part of the small self-explanatory examples that aren't needed. Decided to get rid of this example entirely

src/graphics/render_window.rs
dogunbound: Done

src/graphics/render_window.rs
dogunbound: Done

src/graphics/render_window.rs
dogunbound: Done

CHANGELOG.md
dogunbound: Done

src/graphics/render_window.rs
dogunbound: This is also another small self-explanatory function. I got rid of the usage example for this one.

src/graphics/render_window.rs
dogunbound: Same for request_focus

dogunbound: Same for set_size

crumblingstatue:
Thank you for your work, I'll review it again sometime tomorrow

src/graphics/render_window.rs
```diff
@@ -328,6 +477,7 @@ impl RenderWindow {
     ///
     /// Return the position in pixels
     ///
+    /// # Usage Example
```
crumblingstatue, line 480: There is a stray Usage Example header here, without example

src/graphics/render_window.rs
```diff
@@ -342,6 +492,23 @@ impl RenderWindow {
...
+    /// # use sfml::system::Vector2;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.set_position(Vector2::new(100, 400));
+    /// use std::{thread, time::Duration};
+    /// // You need to wait for the OS the set the window's position before checking
+    /// thread::sleep(Duration::from_secs(1));
```
crumblingstatue, line 509: I think you could lower the sleep to 250 milliseconds. It's reasonable, and doesn't increase test times as much as 1 second.

src/graphics/render_window.rs
```diff
@@ -351,6 +518,23 @@ impl RenderWindow {
...
+    /// # use sfml::system::Vector2;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.set_size(Vector2::new(100, 400));
+    /// use std::{thread, time::Duration};
+    /// // You need to wait for the OS the set the window's size before checking
+    /// thread::sleep(Duration::from_secs(1));
```
crumblingstatue, line 535: Same here as above.

src/graphics/render_window.rs
```diff
@@ -389,6 +572,27 @@ impl RenderWindow {
...
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// # use sfml::window::{ Cursor, CursorType };
+    /// let cursor = Cursor::from_system(CursorType::Arrow);
+    /// if let Some(arrow_cursor) = &cursor {
+    ///     unsafe { window.set_mouse_cursor(arrow_cursor); }
+    /// }
+    /// // You need to ensure the SFML window closes before the cursor's end of life.
+    /// // Doing it the other way around will cause errouneous errors.
```
crumblingstatue, line 592: Errouneous errors? That sounds a bit redundant. Would just "will cause errors" be enough?

src/graphics/render_window.rs
```diff
@@ -419,6 +623,23 @@ impl RenderWindow {
...
+    /// # use sfml::graphics::RenderWindow;
+    /// # // Create a new window with SFML window as name
+    /// # let mut window = RenderWindow::new((800, 600),
+    /// #                              "SFML window",
+    /// #                              Style::CLOSE,
+    /// #                              &Default::default());
+    /// window.request_focus();
+    /// use std::{thread, time::Duration};
+    /// // You need to wait for the OS the set the window's visibility before checking
+    /// thread::sleep(Duration::from_secs(1));
```
crumblingstatue, line 640: 250 ms sleep here as well

crumblingstatue (changes requested):
Thank you, it's looking good. Just a few more things I'd like changed.

src/graphics/render_window.rs
dogunbound: Done.

crumblingstatue:

PR #290, opened 2022-10-27
Support for static linking on Windows and Linux

arm32x:
This pull request adds support for statically linking SFML for Windows (MSVC) and Linux, which can be activated by setting the `SFML_STATIC` environment variable.

This works by linking all dependencies of SFML into `rust-sfml`, which is the correct way to do static linking according to [the SFML FAQ](https://www.sfml-dev.org/faq.php#build-link-static).

I have used this successfully to build statically-linked executables for the `x86_64-pc-windows-msvc` and `x86_64-unknown-linux-gnu` targets. No other targets have been tested and are not likely to work without further changes.

On Linux, SFML must be built from source with the `BUILD_SHARED_LIBS` CMake option set to false, since most SFML packages do not include static libraries.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -4,6 +4,7 @@ fn main() {
     println!("cargo:rerun-if-changed=CSFML");
     println!("cargo:rerun-if-env-changed=SFML_INCLUDE_DIR");
     println!("cargo:rerun-if-env-changed=SFML_LIBS_DIR");
+    println!("cargo:rerun-if-env-changed=SFML_STATIC");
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Thank you, I'll take a closer look tomorrow!

crumblingstatue:
Well.... This requires static libraries for all SFML deps to be present on the system, which is not the case on most systems, and sounds like a pain to prepare for the average user.

But since this doesn't affect dynamically linked SFML, I'm accepting this pull request.

PR #295, opened 2023-01-07
build.rs uses current default toolchain to determine the features enabled

donicrosby:
This PR fixes the build.rs config so that when building for a different target (e.g. Windows on a Linux box) it selects the proper features and dynamic and static libs

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -130,7 +130,7 @@ fn main() {
             .iter(),
         );
     }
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Thank you for your pull request.
I'm currently sick in bed, but I'll review it as soon as I'm feeling better

crumblingstatue:

PR #297, opened 2023-01-11
Added documentation for LD_LIBRARY_PATH

dogunbound:
Had an issue linking after setting the include and lib path as suggested by the rust docs. After I read the C++ SFML docs on linking, I fixed my problem. 

Would be nice to have added documentation just in case we or someone else runs into this in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,10 @@
 //! - `SFML_INCLUDE_DIR`. Set this to the `include` folder of your SFML location.
 //! - `SFML_LIBS_DIR`. Set this to the `lib` folder of your SFML location.
 //!
+//! Linux users may also have to set this environment variable to help your binary find the
+//! shared object files.
+//! - `LD_LIBRARY_PATH`. Set this to the `lib` folder of your SFML location.
+//!
 //! # !! Thread safety warning !!
 //!
 //! rust-sfml strives to be memory-safe, as a Rust library should be, but currently there is no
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -12,9 +12,10 @@
 //! =============
 //! If you get errors about SFML headers not being found, or linker errors, that probably means
 //! SFML is not installed in a global location.
-//! In that case, you can set two environment variables to help rust-sfml find the required files:
+//! In that case, you can set three environment variables to help rust-sfml find the required files:
```
crumblingstatue, line 15: `LD_LIBRARY_PATH` is a linux/*nix specific environment variable to help the dynamic loader find the required dynamic libraries, so putting it here is misleading.

- It doesn't help `rust-sfml` find the required files. It has nothing to do with `rust-sfml`.
- It doesn't work on windows, this is linux/*nix speficic.

Also this technically isn't about linking at compile time, it's about loading dynamic libs at runtime.

crumblingstatue (changes requested):
Should probably go into its own linux specific section, and not grouped in with environment variables that rust-sfml uses.

src/lib.rs
dogunbound: Awesome. Of course, the documentation on the main SFML is superior since it is SFML-specific, but allowing users to fix all their edge cases from one documentation source is more user-friendly. 

It's fixed now

crumblingstatue:
Thanks!

PR #301, opened 2023-01-28
Const default vertex

dogunbound:
Inspiration was to ensure uniformity of the default value. I also wanted to make the default public because it can be handy to build a vertex array with default vertices in const contexts There are other reasons why I wanted to make it public, but they all boil down to making user code cleaner.


crumblingstatue:

PR #303, opened 2023-03-23
Add a try into just in case user wants to handle failed conversions

dogunbound:
I think it would be good to have a non-panic version of `to_rust_string()`


crumblingstatue:
Ideally, we would have our own error type here, that we can keep even if we use a different method, but it would be more concrete than `dyn Error`.
It could be SfStrConvError, or similar.

src/system/string.rs
```diff
@@ -19,6 +19,14 @@ impl SfStr {
     pub fn to_rust_string(&self) -> String {
         self.0.to_string().unwrap()
     }
+
+    /// Convert to a UTF-8 `String` from the Rust standard library.
+    ///
+    /// Returns a `Result` and errors if the string is not valid UTF-32
+    pub fn try_into_rust_string(&self) -> Result<String, Utf32Error> {
```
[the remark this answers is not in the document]
crumblingstatue: In this case I think it would make sense to just have a dedicated error type instead of a generalized error type that's just a string.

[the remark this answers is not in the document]
dogunbound: Alrighty. Made a new error type specifically for UTF32 to UTF8 conversions. I don't think we are doing UTF16 to UTF8 or anything else problematic, so this should work!

src/system/string.rs
```diff
@@ -62,3 +76,64 @@ impl<'a> SfStrConv for &'a String {
...
+        self.inner
+            .as_ref()
+            .and_then(|inner| inner.get(self.index).copied())
+    }
+
+    /// Returns the underlying vector of values which generated the error in the first place.
+    ///
+    /// If the sequence that generated the error was a reference to a slice instead of a [`Vec`],
+    /// this will return [`None`].
+    #[inline]
```
crumblingstatue, line 107: What's with the inline and cfg_attr attributes?

src/system/string.rs
```diff
@@ -62,3 +76,64 @@ impl<'a> SfStrConv for &'a String {
...
+    #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+    #[must_use]
+    pub fn into_vec(self) -> Option<Vec<u32>> {
+        self.inner
+    }
+}
+
+impl From<Utf32Error> for SfStrConvError {
```
crumblingstatue, line 115: Instead of the From impl, it could just wrap Utf32Error, since it has private fields anyway

dogunbound: Could you show me what you mean? 

Also, I see nothing wrong with the from implementation

crumblingstatue:

PR #305, opened 2023-07-03
Implements set_smooth and is_smooth

dogunbound:
New 2.6.x SFML allows for setting the smoothness to fonts. rust-sfml should reflect this ability too.


crumblingstatue:
Thanks, I'll review it later today

dogunbound:
Looks like the builder needs an update. This compiles and tests just fine on my computer. I think the builder itself still uses 2.5

PR #306, opened 2023-07-04
Implement fetching bold kerning

dogunbound:


dogunbound:
Looks like the builder needs an update. This compiles and tests just fine on my computer. I think the builder itself still uses 2.5

PR #309, opened 2023-07-07
Implement isSrgb

dogunbound:


CSFML/src/Graphics/RenderWindow.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Graphics/Rect.h"
 #include "Graphics/Vertex.h"
+#include "SFML/Graphics/RenderTarget.hpp"
```
crumblingstatue, line 3: Same problem as above.

crumblingstatue:
Thanks for the contribution!

PR #312, opened 2023-07-28
Adding a tip for users who try to install rust-sfml using cargo build…

GrimReaper420:
…. Setting the environment variable each time can be tedious and does not comply with rust-analyzer. Using the config.toml file can alleviate some of the headache. This was a real pain for me before I figured this out.


crumblingstatue:
Looks useful, thanks!

Issue #311, opened 2023-07-26
Scancodes are private and thus unusable


andreymal:
I tried to make a typical event handling:

```rust
match event {
    sfml::window::Event::KeyPressed { scan, .. } => {
        match scan {
            sfml::ffi::window::Scancode::Escape => { /* ... */ },
            _ => { /* ... */ },
        }
    },
    _ => { /* ... */}
}
```

...but got an ```error[E0603]: module `ffi` is private```

I guess someone forgot to add `pub use crate::ffi::window::Scancode;` in `src/window/keyboard.rs`

(UPD: and then `pub use self::keyboard::Scancode` in `src/window/mod.rs`)

andreymal:
Oh no, `scan` is not even implemented for `KeyReleased`

crumblingstatue:
This should be resolved by #313. Let me know if you have any problems!

PR #313, opened 2023-08-15
Make Scancode public and add missing scan field to KeyReleased event

NuriYuri:
I also need those. Relates to: https://github.com/jeremyletang/rust-sfml/issues/311


crumblingstatue:
Thanks for your contributions!

Issue #308, opened 2023-07-04
Missing 2.6 features


dogunbound:
I'm just creating an issue for every missing 2.6 feature. This list will just stay here. Whether we implement them or not will be based on personal needs/wants. If someone creates an issue stating that this feature is missing, then we could include it, and document it's implementation here too. 

Missing 2.6 features:
[ ] Implement direction arrow cursors
[ ] Implement sf::Image::saveToMemory

crumblingstatue:
> Not sure how complete it is @crumblingstatue could you verify the completion on scancodes?

I only added the missing Scancode type and union field to restore the broken ABI, the feature is most likely not complete.

dogunbound:
Alright @crumblingstatue 

I'm not sure about how to implement everything else. ill leave it up to you to decide/do them. I don't feel I need them but wanted to at least make rust-sfml do what regular sfml does.

PR #314, opened 2023-08-19
Implement SFML 2.6 direction arrow cursors

NuriYuri:
Hi, little contribution to fix the CursorType enum :)

Since all of those cursors are Linux Specific I can only assess that SizeAll, Cross, Help and NotAllowed works properly (now that they have the correct enum value) and that the newly added cursor behave as explained in the SFML docs.

Relates to: https://github.com/jeremyletang/rust-sfml/issues/308


crumblingstatue:

PR #319, opened 2023-10-17
Implement function to fetch raw texture behind an rc_texture

dogunbound:
This can be useful for many things like allowing the user the use Sprites, Shapes, VertexArrays, etc... with the RcTexture. I found this limitation out recently and realized I kind of needed this function to continue working. I don't see and negative side effects because it is using a borrow, so you can't really use that underlying raw texture as long as the RcTexture is alive.

### DIFF
--- a/src/graphics/rc_texture.rs
+++ b/src/graphics/rc_texture.rs
@@ -127,6 +127,18 @@ impl RcTexture {
         self.texture.borrow_mut().create(width, height)
     }
 
+    /// Get the source texture of an [`RcTexture`]
+    ///
+    /// It let's you temporarily borrow the [`Texture`] being held by an [`RcTexture`].
+    /// This may be useful for many things, for example, using vertex arrays with
+    /// [`RcTexture`].
+    ///
+    /// Returns a [`Texture`]
+    #[must_use]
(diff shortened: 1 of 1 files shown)


dogunbound:
I'm still testing it. I'll let you know once I've tested it. But this looks good. I don't think it won't work.

src/graphics/rc_texture.rs
```diff
@@ -127,6 +127,18 @@ impl RcTexture {
...
+    /// Get the source texture of an [`RcTexture`]
+    ///
+    /// It let's you temporarily borrow the [`Texture`] being held by an [`RcTexture`].
+    /// This may be useful for many things, for example, using vertex arrays with
+    /// [`RcTexture`].
+    ///
+    /// Returns a [`Texture`]
+    #[must_use]
+    pub fn raw_texture(&self) -> &Texture {
+        unsafe { &*self.texture.as_ptr() }
```
crumblingstatue, line 139: I'm not 100% sure if it's sound to just get an immutable reference out of a RefCell like this, but right now I can't find any example where it causes undefined behavior.
But since rc_texture is your domain, I'll just let you reign free.

PR #321, opened 2023-12-25
Ported `sf::Context::getFunction()`

truenotzero:
Ported `sf::Context::getFunction()`:
- Added as `Context::get_function()` 
Low level details:
- Added `sfContext_getFunction()` to CSFML implementation
- Added `sfContext_getFunction()` FFI function 

Possible changes:
Change `Context::get_function()`'s signature, as it currently is: `unsafe fn(&CStr) -> *const std::ffi::c_void`

### DIFF
--- a/CSFML/src/Window/Window.cpp
+++ b/CSFML/src/Window/Window.cpp
@@ -148,6 +148,6 @@ extern "C" uint64_t sfContext_getActiveContextId() {
     return sf::Context::getActiveContextId();
 }
 
-extern "C" void* sfContext_getFunction(char const* name) {
+extern "C" sf::GlFunctionPointer sfContext_getFunction(char const* name) {
     return sf::Context::getFunction(name);
 }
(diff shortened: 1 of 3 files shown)


crumblingstatue:
Hi, thank you for your pull request!

On my system, this produces the following error during building CSFML:

```
CSFML/src/Window/Window.cpp: In function ‘void* sfContext_getFunction(const char*)’:
CSFML/src/Window/Window.cpp:152:36: error: invalid conversion from ‘sf::GlFunctionPointer’ {aka ‘void (*)()’} to ‘void*’ [-fpermissive]
  152 |     return sf::Context::getFunction(name);
      |            ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
      |                                    |
      |                                    sf::GlFunctionPointer {aka void (*)()}
      ```

crumblingstatue:
Returning `sf::GlFunctionPointer` seems to solve it, but not sure if it has any caveats.

truenotzero:
You're absolutely right! That's a bad function declaration, which causes an implicit cast (which is illegal in C++).
Weird that my compiler didn't even warn about it...

I updated the code with your suggestion.

crumblingstatue:

PR #322, opened 2024-02-18
Implement safer pixel data functions

dogunbound:
I believe that it is unnecessary for the user to use unsafe functions when it can easily be implemented with safe functionality. This is an example of doing it safely with a small abstraction. Rust is all about safe code, not unsafe code. Such a simple abstraction should be hidden from the user

Note not inside of commit:
I haven't tested it. I'll test it soon and leave a comment that I have tested it's functionality. 

This will be a breaking change, but I believe it is for the better. People using these functions will have to change their code and decide whether they want to use the unchecked or standard version of the function.

### DIFF
--- a/src/graphics/image.rs
+++ b/src/graphics/image.rs
@@ -1,3 +1,5 @@
+use std::{error::Error, fmt};
+
 use {
     crate::{
         ffi::graphics as ffi,
@@ -186,8 +188,35 @@ impl Image {
     /// This function doesn't check the validity of the pixel
     /// coordinates, using out-of-range values will result in
     /// an undefined behaviour.
-    pub unsafe fn set_pixel(&mut self, x: u32, y: u32, color: Color) {
+    pub unsafe fn set_pixel_unchecked(&mut self, x: u32, y: u32, color: Color) {
         ffi::sfImage_setPixel(self.image, x, y, color)
+    }
+
+    /// Change the color of a pixel in an image
+    ///
+    /// # Arguments
+    /// * x - X coordinate of pixel to change
+    /// * y - Y coordinate of pixel to change
+    /// * color - New color of the pixel
+    pub fn set_pixel(&mut self, x: u32, y: u32, color: Color) -> Result<(), SetPixelError> {
+        let image_size = self.size();
+        if x >= image_size.x {
+            return Err(SetPixelError::XTooLarge {
+                x,
+                width: image_size.x - 1,
+            });
+        }
+        if y >= image_size.y {
+            return Err(SetPixelError::YTooLarge {
+                y,
+                height: image_size.y - 1,
+            });
+        }
+
+        // Since we check for index validity before setting the pixel, it is safe unless the
+        // image has been unloaded, but I doubt you can even do that.
+        unsafe { ffi::sfImage_setPixel(self.image, x, y, color) }
+        Ok(())
     }
 
     /// Get the color of a pixel in an image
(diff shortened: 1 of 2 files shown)


dogunbound:
Ok, i might still be testing this by the time you see this @crumblingstatue 

I'm uploading some debug code, so my cargo build can see it off my branch

dogunbound:
Finally done testing. Here is the test code i used:

```rust
use sfml::graphics::{Color, Image};

fn main() {
    let mut image = Image::new(64, 64);

    for x in 0..64 {
        for y in 0..64 {
            match image.set_pixel(x, y, Color::RED) {
                Ok(_) => (),
                Err(e) => {
                    let error_message = format!("{}", e);
                    println!("{}", error_message);
                }
            };

            match image.pixel_at(x, y) {
                Some(_) => (),
                None => {
                    println!("no pixel at x:{} y:{}", x, y);
                }
            }
        }
    }
}
```

src/graphics/image.rs
```diff
@@ -204,10 +228,29 @@ impl Image {
...
+    ///
+    /// Return the Color of the pixel at coordinates (x, y)
+    #[must_use]
+    pub fn pixel_at(&self, x: u32, y: u32) -> Option<Color> {
+        let image_size = self.size();
+        if image_size.x <= x || image_size.y <= y {
+            return None;
+        }
+
+        // Since we check for index validity before setting the pixel, it is safe unless the
```
crumblingstatue, line 249: You mention setting the pixel here, but we're getting it. Probably copy-paste error.

src/graphics/image.rs
```diff
@@ -293,3 +336,24 @@ impl Drop for Image {
...
+    YTooLarge(u32, u32),
+}
+
+impl Error for SetPixelError {}
+
+impl fmt::Display for SetPixelError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::XTooLarge(passed_in, max) => {
+                write!(f, "x index out of bounds. x:{} max_x:{}", passed_in, max)
```
crumblingstatue, line 352: Perhaps instead of `max_x`/`max_y`, perhaps `width` and `height` would make more sense.

crumblingstatue (changes requested):
Overall I approve of adding safe variants of these functions.
Just a few nitpicks.

src/graphics/image.rs
dogunbound: Fixed!

src/graphics/image.rs
```diff
@@ -186,10 +189,31 @@ impl Image {
...
+        let image_size = self.size();
+        if x >= image_size.x {
+            return Err(SetPixelError::XTooLarge(x, image_size.x - 1));
+        }
+        if y >= image_size.y {
+            return Err(SetPixelError::YTooLarge(y, image_size.y - 1));
+        }
+
+        // Since we check for index validity before setting the pixel, it is safe unless the
+        // image texture has been unloaded, but I doubt you can even do that.
```
crumblingstatue, line 212: Using the word "texture" here is confusing, since this is an image in system memory, not a texture in video memory.

src/graphics/image.rs
```diff
@@ -204,10 +228,29 @@ impl Image {
...
+    /// Return the Color of the pixel at coordinates (x, y)
+    #[must_use]
+    pub fn pixel_at(&self, x: u32, y: u32) -> Option<Color> {
+        let image_size = self.size();
+        if image_size.x <= x || image_size.y <= y {
+            return None;
+        }
+
+        // Since we check for index validity before getting the pixel, it is safe unless the
+        // image texture has been unloaded, but I doubt you can even do that.
```
crumblingstatue, line 250: Using the word "texture" here is confusing, since this is an image in system memory, not a texture in video memory.

src/graphics/image.rs
```diff
@@ -1,3 +1,6 @@
+use core::fmt;
+use std::error::Error;
```
crumblingstatue, line 2: Have you run rustfmt on your code? It should merge these uses.

src/graphics/image.rs
```diff
@@ -293,3 +336,24 @@ impl Drop for Image {
         unsafe { ffi::sfImage_destroy(self.image) }
     }
 }
+
+#[derive(Debug, Copy, Clone)]
+pub enum SetPixelError {
+    XTooLarge(u32, u32),
```
crumblingstatue, line 342: Should probably be named fields for better clarity.` {x: u32, width: u32}`

src/graphics/image.rs
```diff
@@ -293,3 +336,24 @@ impl Drop for Image {
...
+    YTooLarge(u32, u32),
```
crumblingstatue, line 343: Same here as above

crumblingstatue (changes requested):
Some more nitpicks, sorry.

src/graphics/image.rs
dogunbound: Removing the word "texture" is enough here. Fixed!

src/graphics/image.rs
dogunbound: Removing the word "texture" is enough here. Fixed!

src/graphics/image.rs
dogunbound: Fixed

src/graphics/image.rs
dogunbound: Fixed

dogunbound:
Cargo fmt caught something in `src/window/context.rs` also. I added it as a seperate commit. Thought I might as well include it into this PR since it was already up.

src/graphics/image.rs
dogunbound: fixed. It actually caused other changes in seperate files. made a separate commit in this pr to address them.

PR #323, opened 2024-02-21
Fix clippy errors

dogunbound:
Clippy was complaining that #[must_use] MUST be implemented for this function.

It was also complaining that a # Safety documentation has yet to be written.

The only part I believe that needs a code review is the message for the # Safety documentation. I've never used this function, so I am unsure what it does. I did my best reading off the context clues what the official SFML documentation, and usages I could find.

### DIFF
--- a/src/window/context.rs
+++ b/src/window/context.rs
@@ -69,8 +69,14 @@ impl Context {
     /// Get the address of an OpenGL function.
     /// # Arguments
     /// * name - Name of the function to get the address of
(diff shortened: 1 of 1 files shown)


crumblingstatue:
Thank you!

I ended up making `Context::get_function` safe, since it just returns a pointer.
If the user wants to dereference that pointer, they will have to use `unsafe` anyway.

I put `#[must_use]` on it as well.

Issue #316, opened 2023-09-09
Failed when compiling CSFML


Cg1340:
It looks like an import is needed ``<cstdint>``, but it is not imported.

OS: Windows 11 x64
C&C++ Compiler: Mingw-w64 with version 13.1.0 of gcc
rustc: 1.72.0

```
PS D:\Code\musicplayer-continue> cargo build
   Compiling sfml v0.21.0
The following warnings were emitted during compilation:

warning: Custom SFML include dir: D:\SFML-2.6.0\include
warning: CSFML/src/Audio/Sound.cpp:74:60: error: 'int64_t' has not been declared
warning:    74 | extern "C" void sfSound_setPlayingOffset(sf::Sound *sound, int64_t timeOffset) {
warning:       |                                                            ^~~~~~~
warning: CSFML/src/Audio/Sound.cpp:103:12: error: 'int64_t' does not name a type
warning:   103 | extern "C" int64_t sfSound_getPlayingOffset(const sf::Sound *sound) {
warning:       |            ^~~~~~~
warning: CSFML/src/Audio/Sound.cpp:3:1: note: 'int64_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
warning:     2 | #include <SFML/Audio.hpp>
warning:   +++ |+#include <cstdint>
warning:     3 | #include <cstddef>

error: failed to run custom build command for `sfml v0.21.0`

Caused by:
  process didn't exit successfully: `D:\Code\musicplayer-continue\target\debug\build\sfml-66243c0e90a514ed\build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-changed=CSFML
  cargo:rerun-if-env-changed=SFML_INCLUDE_DIR
  cargo:rerun-if-env-changed=SFML_LIBS_DIR
  cargo:rerun-if-env-changed=SFML_STATIC
  cargo:warning=Custom SFML include dir: D:\SFML-2.6.0\include
  TARGET = Some("x86_64-pc-windows-gnu")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-pc-windows-gnu")
  cargo:rerun-if-env-changed=CXX_x86_64-pc-windows-gnu
  CXX_x86_64-pc-windows-gnu = None
  cargo:rerun-if-env-changed=CXX_x86_64_pc_windows_gnu
  CXX_x86_64_pc_windows_gnu = None
  cargo:rerun-if-env-changed=HOST_CXX
[...]

PR #328, opened 2024-06-09
Fix compiling error on Windows MinGW

wanzzhehe:
Add `<cstdint>` header file to `CSFML/src/Audio/Sound.cpp` to avoid compiling error.


wanzzhehe (commented):
> Add `<cstdint>` header file to `CSFML/src/Audio/Sound.cpp` to avoid compiling error. 

I found the issue #316 and I have the same problem. But there is no any PR to fix it. So I just open up this PR.

crumblingstatue:

Issue #179, opened 2017-11-19
Figure out what types are safe to implement `Send` for


crumblingstatue:
`Image` might be one such type, so you could load images on a separate thread, and send them to the main thread.

crumblingstatue:
#177 relates to this, `Window` doesn't implement `Send`.

qwandor:
Relatedly, is there any reason why `Sound` isn't both `Send` and `Sync`?

PR #331, opened 2024-11-21
Make various types Send and Sync

qwandor:
`FBox` should be `Send` and `Sync` if its contents is, the same as `std::boxed::Box`.

`Music`, `Sound`, `SoundRecorderDriver` and `SoundBufferRecorder` can also be both `Send` and `Sync` as far as I can see. They can be `Send` because nothing about them is tied to a particular thread, and `Sync` because the underlying C++ implementation uses atomics for shared state they may access.

See #179.

### DIFF
--- a/src/audio/capture.rs
+++ b/src/audio/capture.rs
@@ -96,6 +96,15 @@ pub struct SoundRecorderDriver<'a, R: 'a> {
     recorder: &'a mut R,
 }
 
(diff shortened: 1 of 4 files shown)


crumblingstatue:
I'm not an expert on thread safety issues, but on the surface, this looks good to me. Thank you for your contribution!

PR #332, opened 2024-12-22
Add -ObjC link flag to fix OSX crashes

kesyog:
Add `-ObjC` link flag when building on OSX. As noted in https://github.com/SFML/SFML/issues/2920, this is required when statically linking against SFML since it contains Objective-C source files. Without this flag, compiling and linking succeeds but binaries will crash at runtime.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -303,6 +303,9 @@ fn main() {
         if feat_audio {
             unix_audio_link_support_libs();
         }
+        // SFML contains Objective-C source files on OSX
+        // https://github.com/SFML/SFML/issues/2920
+        println!("cargo::rustc-link-arg=-ObjC");
     } else {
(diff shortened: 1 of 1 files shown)


kesyog:
Unfortunately, it seems that due to https://github.com/rust-lang/cargo/issues/9554, this link flag isn't propagated to downstream users, but this change at least lets me run the examples within this crate. 

Until a workaround is found, perhaps we should document in the README or Wiki that downstream users should add this to their build scripts?

crumblingstatue:
I have no way to test OS X personally, but CI passes, so looks good to me.

> Until a workaround is found, perhaps we should document in the README or Wiki that downstream OSX users should add this to their build scripts?

Feel free to edit the wiki, or submit a documentation pull request.

PR #334, opened 2025-01-28
Update cmake dependency to 0.1.53

kesyog:
This cmake-rs release contains a fix that speeds up cmake builds on OSX.

See https://github.com/rust-lang/cmake-rs/releases/tag/v0.1.53


crumblingstatue:
Hi, sorry for the late response, I totally forgot about this :sweat:

Issue #339, opened 2025-04-03
CMake 4.0 breaks on libogg (libflac-sys)


crumblingstatue:

crumblingstatue:
One workaround is to set the environment variable `CMAKE_POLICY_VERSION_MINIMUM=3.5` when building `libflac-sys`.
Perhaps a patch could be submitted upstream for their build script to set it.

kesyog:
I put up https://github.com/mgeier/libflac-sys/pull/39 to update the flac and ogg libraries in `libflac-sys`, as both libraries' CMake build files have been fixed to work with CMake 4.0 in their latest versions

kesyog:
Another option is to add a feature flag to rust-sfml to dynamically link flac and ogg by disabling `libflac-sys`'s default features and changing the link type from static to dylib in our build.rs.

PR #341, opened 2025-04-07
Allow dynamic linkage of FLAC and ogg

kesyog:
* Allow libFLAC and libogg to be dynamically linked rather than built from source by libflac-sys and statically linked.
* The static linking of these two libraries is maintained as the default behavior but one can choose to dynamically link them by opting out of the new `build-flac-ogg` feature flag, which is on by default.

Tested on OSX and Debian-ish Linux by:
1. Running the pong example with dynamic linking of libogg and libFLAC: ` cargo run --example pong --no-default-features --features=audio,graphics,build-flac-ogg`
2. Running the pong example with static linking of libogg and libFLAC:  `cargo run --example pong`
    * Used https://github.com/mgeier/libflac-sys/pull/39 to allow it to build with CMake 4.0 on OSX
    * Tested with an older version of CMake and vanilla libflac-sys on Linux

This also works as a workaround for  #339 until libflac-sys is updated (https://github.com/mgeier/libflac-sys/pull/39).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Install deps
       run: |
        sudo apt-get update
-       sudo apt-get install libpthread-stubs0-dev libgl1-mesa-dev libx11-dev libx11-xcb-dev libxcb-image0-dev libxrandr-dev libxcb-randr0-dev libudev-dev libfreetype6-dev libglew-dev libjpeg8-dev libgpgme11-dev libsndfile1-dev libopenal-dev libjpeg62 libxcursor-dev cmake libclang-dev clang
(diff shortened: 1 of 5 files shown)


crumblingstatue:
Thank you for your PR!
Can you look into why CI is failing for windows and OS X? (cannot find libFLAC)
Is dynamic linking getting enabled by default?

kesyog:
I had to modify the actions files, as building with `--no-default-features` and the `audio` feature will now try to dynamically link FLAC and ogg. I have CI building OSX and Linux with both methods and Windows building via static linking only as I'm not sure how to set that up to work.

Note: I did run into an unrelated compilation issue with OSX after an update last night, as the currently-vendored SFML C++ code contains non-standard std::string usage. I'll create a separate issue for that.

crumblingstatue:
Wouldn't it be possible to have a `dylink-flac-ogg` feature instead, so no existing configuration gets broken, and the CI script doesn't have to be changed either?

kesyog:
The libflac-sys crate exposes `build-flac` and `build-ogg` features. Since features are additive, I don't think there's a way to enable those features by default and only disable them if a new `dylink-flac-ogg` rust-sfml flag is enabled. But let me know if you have ideas for how to structure the flags to make that work. Or if you think this approach is too messy for the benefit.

crumblingstatue:
Ah, I see, thank you for the information.
I'll have to think a bit more about this. 
I'll do a proper review tomorrow.

crumblingstatue:

PR #350, opened 2025-10-20
Add ownership for InputStream and Music.

stopmind:
Add methods for `InputStream` and `Music` that allow them to own their sources.
It will make it possible to use them without lifetimes with custom source.

### DIFF
--- a/src/audio/music.rs
+++ b/src/audio/music.rs
@@ -66,7 +66,7 @@ impl<'src, S> Music<'src, S> {
     }
     /// Create a new `Music` by "opening" it from a stream, with ownership.
     ///
-    /// See [`Self::open_from_stream`].
+    /// See [`Self::open_from_stream_owned`].
     pub fn from_stream_owned(stream: InputStream<'src, S>) -> SfResult<Self> {
         let mut new = Self::new()?;
         new.open_from_stream_owned(stream)?;
--- a/src/audio/music.rs
+++ b/src/audio/music.rs
@@ -9,8 +9,7 @@ use {
     },
     std::{
         ffi::{CString, c_uint, c_void},
-        io::{Read, Seek},
-        marker::PhantomData,
+        io::Empty
     },
 };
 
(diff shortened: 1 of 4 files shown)


src/audio/music.rs
```diff
@@ -48,55 +47,52 @@ unsafe impl Send for Music<'_> {}
...
         Ok(Self {
             music: unsafe { ffi::audio::sfMusic_new() },
-            _stream: PhantomData,
+            stream: None,
         })
     }
-    /// Create a new `Music` by opening a music file
-    ///
-    /// See [`Self::open_from_file`].
-    pub fn from_file(filename: &str) -> SfResult<Self> {
```
dogunbound: Why does from_file need to be removed? I'm not well versed in the music stuff, so some explanation may be needed.

dogunbound (approved):
Ah I see. So we remove from_file in the lifetime requirement section because it doesn't make sense to borrow from a file. That sort of abstracts away some of the fact that we are running on an underlying API.

Same with from_memory. That's good.

A bit of detail why in the commit message would be nice, but other than that looks fine. 

ALso I'm kind of dumb when it comes to this music stuff, so I'm sort of trusting you. @crumblingstatue his added example does run and this does compile. lgtm.

dogunbound:
@crumblingstatue maybe we should have it run checks all the time that way someone can see the builder failures?

stopmind:
Sorry, I forgot to specify required features for the new example.

dogunbound (approved):
assuming builder passes.

crumblingstatue:
> @crumblingstatue maybe we should have it run checks all the time that way someone can see the builder failures?

I don't think I have permission to enable that

dogunbound:
> > @crumblingstatue maybe we should have it run checks all the time that way someone can see the builder failures?
> 
> 
> 
> I don't think I have permission to enable that

Damn. Maybe there is some way to change who has control of the rust cargo package and we can make a new repo... or maybe there is a way to change the repo pointer in the cargo database. Idk how this works.

PR #352, opened 2026-08-06
Add PartialEq to Vertex

dogunbound:
It would be nice for unit testing if Vertex already derived PartialEq. I would't imagine most prod code to use this, but unit testing will.


dogunbound:
@crumblingstatue A merge of this would be nice. Can't see why not.

